PR #189, opened 2020-06-01
New Clock Tree and Upgraded HAL

stephenhensley:


stephenhensley:
This also now has a HAL upgrade. -- The PLL's fractional component was not being properly calculated between the old HAL version and the software generating the code.

This prevented us from effectively recalculating all of the clocks for new 480MHz max speed.

Given we were on a pretty early version of the HAL (w/ respect to the H7 seriest) probably a benefit to upgrade anyway.

Before merging I'm going to give a lot of stuff a solid test, specifically SDMMC which may have had a bug either reintroduced or fixed between versions.

The USB Middleware may also have an upgrade, which may (but hopefully) doesn't depend on any of the changes to the HAL. So I'm going to look at that, too.

PR #192, opened 2020-06-03
Doc

beserge:
Still some problems i.e. autotoc_md throughout pdf. I suspect the boilerplate /**< # */ comments cause this.


beserge:
Autotoc problem fixed, now uses & as boilerplate char

stephenhensley:
First pass looks pretty good.

Looks like the old `/**< # */` boilerplate got missed in a few spots.

Not sure how I want to handle the boilerplate yet. As it makes the resulting documentation look very unfinished (though, it is). 

For now I'd just find the boilerplate you missed, and switch them to the new standard so that we can easily strip them out if we want. or replace them with a generic snippet.

a fun one-liner for that would be:

`find . -name "<filename pattern>" | xargs sed -i "</findpattern/replace/>"`

The overall organization is a little different than how daisysp turned out (it is a bit more complicated, but I think the mixed c/c++ makes it a bit stranger as well).

I'm going to look into the Doxygen file stuff and see how we can get it to appear a bit more organized before we start to deploy this.

stephenhensley:
I think there is still a lot of work to be done wrt the contents of the documentation. But this is a huge step forward for the time being.

We can gradually go through and clean this up as we work on the files in question now.

This PR doesn't include deploying the docs, and that will come along soon with CI, etc.

PR #205, opened 2020-06-19
Warning fixes and misc features

stephenhensley:
Fixes unused variable warnings in a few files, adds utility functions for things that were just missing (i.e. basic MIDI output, Gate Input State, etc.)


stephenhensley:
there are some style issues, and such. but we'll be doing the doc style stuff next  so it'll all get caught with those PRs

PR #207, opened 2020-06-20
changed per_gpio function arguments to const for more flexibility

TheSlowGrowth:
I was just trying to something like this:
```
class ButtonReader
{
public:
    ButtonReader()
    {
        dsy_gpio_init(&leftBttnGpio_);
    }
    // ...
private:
    static constexpr dsy_gpio leftBttnGpio_ = { { DSY_GPIOA, 0}, DSY_GPIO_MODE_INPUT, DSY_GPIO_NOPULL };
};
```
But it doesn't work because dsy_gpio_init  takes e non-const pointer...
So I though I'd change it to a const pointer and make a pull request.

### DIFF
--- a/src/per_gpio.c
+++ b/src/per_gpio.c
@@ -2,7 +2,7 @@
 #include "per_gpio.h"
 #include "util_hal_map.h"
 
-void dsy_gpio_init(dsy_gpio *p)
+void dsy_gpio_init(const dsy_gpio *p)
 {
     GPIO_TypeDef *   port;
     GPIO_InitTypeDef ginit;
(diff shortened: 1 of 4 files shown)


TheSlowGrowth:
I changed the function arguments in `AdcChannelConfig` to const ptrs to be in line with the rest of libDaisy. It also allows to supply `nullptr` for unused mux pins.

I moved the mux pin initialisation from `AdcChannelConfig` to `AdcHandle`. Only the required pins are initialised. I find this more consistent, I hadn't expected `AdcChannelConfig` to be anything more than a POD struct with some fancy setter functions. Also, you could manually write to the member functions of `AdcChannelConfig` and your pins wouldn't be initialised.

TheSlowGrowth:
I don't have hardware yet, please test the changes before merging.

stephenhensley:
Since this relies on the gpio one (that I already fixed docs for) I can just get both of these in a PR and merge. (I just brought in these changes).

I'm also going to change `muxChannels` to `mux_channels` to match the rest of the library's style, and fix a line in `daisy_field.h` that uses the mux.

stephenhensley:
As mentioned in #209 I'll lump these two into a single PR with a few minor changes.

Needed this one to test the ADC one. So easiest to merge them at once on my end.

I'll close this once its merged.

PR #209, opened 2020-06-21
ADC improvements

TheSlowGrowth:
Some improvement suggestions for per_adc.

I moved the mux gpio init into `AdcHandle` and out out of `AdcChannelConfig`. The latter seems more of a POD type with fancy setter functions, so I find it more consistent, if it doesn't actually DO anything.
Also, you could manually write the members of `AdcChannelConfig` and pass them to `AdcHandle` and then your mux GPIOs wouldn't be initilized.

I don't have hardware yet. Please test before merging.


stephenhensley:
Hahahahaha I was just about to make a new PR for this and the constPtrs

PR #210, opened 2020-06-21
Added non interleaved audio callback to Daisy Seed

TheSlowGrowth:
Adds non interleaved audio callbacks to the Daisy Seed.
I know the whole audio system is going to be overhauled... But I'd still like to work with non interleaved audio right away.

I haven't tested this on the hardware, my Seed is still stuck in customs...


TheSlowGrowth:
Should be fine now

Issue #160, opened 2020-03-12
USB CDC Transmit works for either port, but won't work for both simultaneously


stephenhensley:
e.g.
this doesn't work (only the first one called actually transmits):
```C++
dsy_system_delay(500);
sprintf(buff, "Tick:\t%d\r\n", hello_count);
hw.usb_handle.TransmitExternal((uint8_t*)buff, strlen(buff));
sprintf(buff, "Tick:\t%d\r\n", hello_count);
hw.usb_handle.TransmitInternal((uint8_t*)buff, strlen(buff));
hello_count = (hello_count + 1) % 100;
```
while either of these work fine on:
```C++
dsy_system_delay(500);
sprintf(buff, "Tick:\t%d\r\n", hello_count);
hw.usb_handle.TransmitInternal((uint8_t*)buff, strlen(buff));
hello_count = (hello_count + 1) % 100;
```

```C++
dsy_system_delay(500);
sprintf(buff, "Tick:\t%d\r\n", hello_count);
hw.usb_handle.TransmitExternal((uint8_t*)buff, strlen(buff));
hello_count = (hello_count + 1) % 100;
```

the redundant `sprintf` is there because at first I thought the transmit function may have been consuming the buffer when transmitting, but it doesn't.

jcolicchio:
Issue is in `usbd_cdc.c`, `h_cdc_fs` or `h_cdc_hs` is being used by both USB ports. We can declare a single `h_cdc[256]` and use `pdev->id` to select which handle to use during init, or we could declare `h_cdc[256]` as an array of `USBD_CDC_HandleTypeDef *`, and malloc/free handles on `Init` and `DeInit`

I can draft a PR with one of these fixes for review and iteration

stephenhensley:
fixed in #213

PR #213, opened 2020-06-29
Bugfix/simultaneous transmit

jcolicchio:
Fix for #160 

This PR attempts to fix an issue where multiple USB ports shared global state internally, and would be unable to transmit simultaneously.

It also returns the transmit result, if we think that'd be handy to return.

Finally, this PR implements line_coding for external port; I wasn't sure if this was required

### DIFF
--- a/Middlewares/ST/STM32_USB_Device_Library/Class/CDC/Src/usbd_cdc.c
+++ b/Middlewares/ST/STM32_USB_Device_Library/Class/CDC/Src/usbd_cdc.c
@@ -94,10 +94,6 @@
   * @}
   */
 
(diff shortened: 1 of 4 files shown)


stephenhensley:
this looks good. I'm going to test it with my VS build tomorrow and make sure whatever malloc related issue I experienced is gone, and then I can merge it.

stephenhensley:
@jcolicchio oops I prematurely checked the ready for review button..

You may want to just rebase with master again, because this one's showing every file changed again/also.

jcolicchio:
@stephenhensley Done! PR is looking good 👍

PR #215, opened 2020-06-30
CI

stephenhensley:


stephenhensley:
tabs and spaces aren't updating with my clang-format installation at home.. calling it a night. 

The actual lib and docs are passing, and doc deployment is set up for master branch. So should be good to go once style is passing.

TheSlowGrowth:
Can we include some sort of check for the daisyExamples? 
I'm wondering how we could make sure that daisyExamples and libDaisy stay consistent to each other without managing two simultaneous pull requests for each change.
I'm starting to think the only way would be to integrate the examples directly into libDaisy. What do you think?

stephenhensley:
Agreed, it is a bit of a pain.

Keeping the three repos separate is nice for a few reasons. Many of the reasons are unique to each repo (for example daisysp being able to move to PCs other platforms a lot easier than libdaisy).

For examples this is a conundrum because most of the DaisyExamples examples require both libdaisy and DaisySP.

having CI on DaisyExamples will be able to essentially check all three repos since they're already submodules, but it doesn't help fix the issue of requiring parallel PRs for changes.

I'm definitely open to solutions, but it doesn't feel like moving all of the examples here would fix much. We'd end up with a similar issue with parallel changes to DaisySP at that point.

Combining all three repos would fix this (they were originally all one repo), but I could see this becoming very unwieldy when the libraries get larger.

TheSlowGrowth:
Right now, daisyExamples is more of a beginners guide than a testing framework. Maybe we can separate both concepts a little and split things up.

We could use daisyExamples as the introductary example pool with "macroscopic" examples that show how things would integrate to be used for making music.
At the same time, daisySP and libDaisy get their own examples (& maybe even automated tests) that are as barebones as possible with no dependencies on anything other than their own repos. Those examples would be more "microscopic" and help identify problems with the code.

When working on libDaisy or daisySP, developers would only check the examples and tests inside those repos. At a later stage, the finished changes will be integrated into daisyExamples via submodule updates. At this point, someone would have to resolve any issues there manually.

Maybe this would make everything even more complicated. Idk. It certainly would reduce the overhead of dealing with two PRs at once. But it would also mean that submodule updates in daisyExamples will become more and more daunting, the more examples are added over time.

stephenhensley:
It's a tricky issue because it needs to balance ease of use, and ease of development with a pretty big divide between what makes each of those easy.

DaisySP is the trickiest because it depends on libdaisy. Otherwise, just adding dedicated libdaisy examples (starting with many of the non-dsp seed/ examples in DaisyExamples) as a test-suite of sorts would be a good way to get most of the way there on this repo. It would at least drastically reduce the number of examples that would need any changes.

The exceptions would be the breakout board files (`daisy_patch.h`, etc.)

----

Another thing is, right now the API for the higher level aspects of the library is very fluid, but that will also be the first part to settle a bit, and then the guts can get worked out without changing user code as much. That will also have a large effect on how often DaisyExamples would need to change parallel to libdaisy.

PR #220, opened 2020-07-04
Feature/more adc improvements

TheSlowGrowth:
builds on top of #219 

Some small improvements to the ADC driver


TheSlowGrowth:
Hmm, my local version of clang-format doesn't generate the same formatting as the travis one. Sigh. I'll repair this tomorrow.

PR #221, opened 2020-07-04
Feature/I2C with DMA

TheSlowGrowth:
Builds ontop of #220 

Changes:
- dsy_i2c_config is used as a dumb POD configuration struct. I used a different way of storing the pin configuration that I find more elegant. Please review.
- I configured DMA1 Stream 6 for I2C 1-3. As all three share the same DMA stream, I had to implement a scheduler that queues and executes I2C transmissions when the DMA is currently in use.
- I added a callback for DMA transmissions. This helps with device drivers that want to do multiple transmissions in a row automatically (e.g. LED drivers)

Things to test further:
- I have only a single I2C slave device to test the scheduler. It works fine for queuing transmissions for the same I2C peripheral, though. If you have hardware with multiple I2C slave devices, please check this as well.
- Error handling: I added the I2C error ISR and it works when sending data to a non-existent device address. I'm not sure if additional error handling would be needed.


TheSlowGrowth:
The build fails due to `dev_leddriver.h` - I replaced this in #222. Should I add a fix for the old `dev_leddriver.h` to the end of this PR?

Also: I'll fix the formatting tomorrow. My local clang-format produces a different output than the travis one.

stephenhensley:
confirmed with the changes to leddriver this works, and tests okay.

I see this all seems fixed on the next PR. So I'll look ahead to that one.

TheSlowGrowth:
> Its funny seeing this get fleshed out in C

Yes, I though about this as well. I'm not sure what's the best way to proceed here. I tried to stay close to the original design, but I'd be totally in for a C++ version of this. Peripherals are kind of global, and there's this init stuff going on behind the scenes where the different platforms setup the pins in their init functions. All of this wouldn't work in the same way if we moved it to C++.
So basically I shied away from a C++ version to not break existing workflows.

If you feel confident moving this to C++, I'd totally be in for a round of refacturing. Let me know what you would prefer.

stephenhensley:
>Peripherals are kind of global, and there's this init stuff going on behind the scenes where the different platforms setup the pins in their init functions. All of this wouldn't work in the same way if we moved it to C++.

So I sort I have a sort-of compromise set-up on the UART peripheral that could probably apply here, it doesn't help much with the extra function calls/messy-ness you were describing, but it does work, and manages to keep what users see pretty clean.

Basically:
* the header file has the interace for interacting (there's no data in the class at the moment)
* the cpp file has a  struct `uart_handle`, and a single static instance that has the HAL stuff, and internal state data (I'm sure some of this could be moved to the UartHandler class (also the naming could be revised to be a little less confusing).
* the two relevant IRQHandlers are implemented at the bottom of the file in and `extern "C"` block

Some of the mechanisms within `uart_handle` could be moved to the class, and a pointer to the class could be added instead. That would allow you to call member functions from the callbacks and such. 

It's definitely not idiomatic c++, but its something. Definitely interested in your thoughts.

TheSlowGrowth:
Stephen, thanks for your thoughts. I have refactored everything to C++ but it's not yet functional. 
I went a similar route - basically I use a pimpl pattern, except that the I2CHandle doesn't own its implementation object, because we have 4 of them only and they're allocated statically. The I2CHandle is basically just a wrapper around the (private, not exposed) implementation class that lives in the cpp file only.
I hope to be able to finish it before the weekend. We'll see. The upside is that I can reuse the programming model for the SPI driver later.

stephenhensley:
Glad to hear it. 

I really need to read up on more idiomatic C++ design patterns now that I'm using it all the time.

With this sort-of architectural stuff I think its worth spending a little more time working out how we want it to look on one thing, rather than push ahead and have to change 10 things multiple times later, so no rush! I haven't had time to check why the multiple LED drivers weren't working on my end yet anyway.

TheSlowGrowth:
@stephenhensley 
I have reworked it again to a Pimpl-style C++ interface. The build fails, but for the same reason as before: The old leddriver can't work with the C++ I2C driver. The next PR fixes this be reworking the led driver as well.

stephenhensley:
Since these changes are included in #222 are we okay to close this PR? I'll be testing/reviewing that PR since it has the functional bits to test this.

TheSlowGrowth:
yes, that's fine. Closed.

PR #222, opened 2020-07-04
Feature/reworked led driver

TheSlowGrowth:
builds on top of #221 

I reworked the LED driver for PCA9685 devices.

Improvements:
- included gamma correction
- multiple independent drivers can be used at the same time by instantiating multiple instances of the class
- no fixed limit to the number of chips anymore
- using the callbacks from the new I2C driver, all chips are now automtically updated in a row.
- double buffered operation with optional data persistence
- led switch-on-times are staggered to reduce noise and PSU ripple


TheSlowGrowth:
I'll fix the formatting tomorrow. My local clang-format produces a different output than the travis one.

stephenhensley:
Seems like the stuff I mentioned in #221 are fixed here.

I like the renaming of the LEDDriver to be device specific, this is a welcome change. The templating is handy, however having to bring `util_hal_map.h` into a device header means that the entire HAL is now being included in any libdaisy code. I would like to avoid this.

Actually it looks like you're not using any of the hal_map functions, but are using HAL delay, this can be replaced with dsy_system_delay, and the hal won't have to be included anymore.

I haven't tried debugging it yet, but the GeneralFunctionTest on the Petal doesn't work anymore (it did work with a modified #221 ) The LEDs assume their first position and then hang indefinitely. I have another board with only a single LED driver that may be a little easier to debug, and can look into a bit more later.

TheSlowGrowth:
I have found the culprit for the broken Petal test. 
I have also removed the HAL include.
@stephenhensley please have a look

stephenhensley:
Just on a first pass, this looks good! I'll be able to test, and dig in a bit deeper later today.

I have a few things that are libdaisy-based with WM8731 on them still so I'll probably have to rework that driver to work with the changes made here, but that's not such a big deal, and those projects can work off the earlier commit they're still on for the time being as well.

TheSlowGrowth:
rebased to current master + some doc cleanups.

stephenhensley:
So for some reason the leddriver still isn't working on DaisyField/DaisyPatch (both with multiple drivers).

I have one board that has a single PCA9685 to make sure its something to do with multiple units, but my Visual Studio  build won't work until I fix up the two Codec files  that use I2C or remove them. So it'll be a little before I can debug it a bit.

stephenhensley:
Okay, took a few to get my single LED Driver project working with the new code, but it seems to work totally fine. Which leads me to believe its still something that's unique to using multiple LED Drivers.

Just a few other review notes while I check out everything else:

* camel-case function names don't fit with the style used within the rest of the library (i.e. `swapBuffersAndTransmit` should be `SwapBuffersAndTransmit`, etc.). 
* I like the new config for I2C a lot.
* The PCA9685 can run at 1MHz so the Field/Petal can have that as the I2C config, but it does work either way.
* It might have been my code since its the only project I have with a single LED driver, and its still in pretty early stages, but the LEDs all the way on seemed to be visibly flickering (even once I increased to 1MHz). Again, this could be user-end code writing the LEDs to slow or something else I didn't check thoroughly enough.
* Saw the usage of a decent amount of `enum class` vs plain `enum` Just curious on when/why you're using them in this context. I remember reading about this recently, but its not fresh in my memory, and it piqued my interest being used here.

stephenhensley:
Also (sorry for comment spam), since I'm realizing you don't have a way of testing the multiple LED driver scenario its worth mentioning that it does seem to do the first transmit, as the LEDs update to their initial update after a reset or power cycle, but don't continue to update after that.

TheSlowGrowth:
1. I fixed the camel case function names. Sorry, still getting used to the style guide and finding it very unfamiliar sometimes :-D
2. I have found the signals to look very bad on my board and assumed that the PCA has a low input impedance om its I2C pins. Hence the lower speed. I'll check that again.
3. I can reproduce the flickering, not sure what's going on. I'll dive in deeper later today.
4. enum class places the names not in the parent namespace (like enum does) => less clutter. It forces you to write the whole thing, e.g. `Config::Peripheral::I2C_2` instead of `Config::I2C_2` which I find more clean in most cases. Also casts to int must be explicit, which I find helpful in avoiding bugs, but YMMV.

stephenhensley:
1. No problem, I totally get it. Its a lot different than what I was originally used to, as well. 
2. Interesting, what values are your pullup resistors? That can have a pretty drastic effect on the rise times. 
3. Okay, good to know. 
4. Ah, makes sense, and seems fitting especially for config items.
5. Yeah, I'm planning on checking it out today. Shouldn't take long to narrow down where the issues coming from.

TheSlowGrowth:
2. I went as low as 1k to get acceptable rise times. I should check my hardware, but it seems to work fine now. I changed the field and petal code to use 1MHz now.
3. The flickering is gone now.

TheSlowGrowth:
@stephenhensley Have you had a chance to look at the dual driver issue?

stephenhensley:
@TheSlowGrowth Sorry, not yet. The past week has been crazy busy on my end. I should have time a little later this week. 

Sorry for the delay on this.

upbeat27:
It is worth noting that 1MHz I2C should not be used if other devices on the bus are only 400kHz, as they might not work correctly (even if the speed is changed to 400kHz when talking to those devices).

stephenhensley:
So I don't know what's different this time, but its totally working with multiple LED drivers now. It could have been something unrelated in the program I was testing at the time. 

I'm going to test on another piece of hardware as well since I have a few that use multiple Drivers, and then look over everything. But provided I don't find anything we'll get this merged shortly!

TheSlowGrowth:
ouch, I just saw that there was some '#if 0' stuff left in the header. Sorry!
I removed it.

stephenhensley:
No problem! And I think I merged it just before you removed it. 

Feel free to open another PR for it if you want.

PR #223, opened 2020-07-05
Renormalize after changes to `.gitattributes`

jcolicchio:
`.gitattributes` was recently modified in 7a05638, but only a subset of affected files were renormalized. This PR is the result of running `git add --renormalize -A` according to [this section](https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#refreshing-a-repository-after-changing-line-endings), in order to renormalize all affected files so that their line endings reflect the settings in `.gitattributes`

To verify behavior, we can check out this branch on various systems, modify source files, and confirm whether the entire file is marked changed. If all is working as expected, only the modified lines should show up in the diff.


jcolicchio:
@stephenhensley I've created this PR to address the issue I've been experiencing with line endings on OSX. It should work for Windows users as well, but to be safe it'd be nice to verify we're not thrashing

TheSlowGrowth:
Oh yes, please include this PR. It's a pain to work right now. Now almost all files show up as "changed" with no way of resetting it because git keeps overwriting the line endings.

jcolicchio:
> Oh yes, please include this PR. It's a pain to work right now. Now almost all files show up as "changed" with no way of resetting it because git keeps overwriting the line endings.

@TheSlowGrowth for a short-term fix to reset git, you can remove all but the first line of `.gitattributes`, at which point git will only report `.gitattributes` as changed; you can then `git checkout .gitattributes` and the changes should no longer register 👍

stephenhensley:
@jcolicchio @TheSlowGrowth 

I'll be able to jump on these PRs today, and this is the first one I planned on reviewing/merging! 😄

jcolicchio:
> @jcolicchio @TheSlowGrowth
> 
> I'll be able to jump on these PRs today, and this is the first one I planned on reviewing/merging! 😄

Awesome! My simultaneous transmit PR has already been rebased onto `renormalize`, so hopefully after merging this branch there should be minimal code changes and no conflicts from simultaneous transmit -> master 🤞

stephenhensley:
Looks good!

PR #224, opened 2020-07-07
Feature/folders

TheSlowGrowth:
This builds on top of #222 

As discussed earlier, with a growing library the current file tree becomes more difficult to navigate. I moved all files into corresponding subdirectories, removing the file prefix and turning it into a directory name.

I adjusted include paths and the makefile, and I hope there's not something more that I have missed.


stephenhensley:
I know this is kind of on hold pending #221 etc. but do you have any reason for using relative include paths, instead of adding the directories to the `-I` list? (e.g. `#include "per/spi.h"` vs `#include "spi.h"`)

I've seen things do it both ways, but if we wanted to put the daisy_seed, daisy_pod, etc. files in a boards folder it would make it a bit less intuitive for users to access them. So I can see why you left in them in `src/` 

Curious on your thoughts.

TheSlowGrowth:
I had removed the prefix from the file names, so `per_spi.h` became
`spi.h` and I figured it would be good to leave the prefix in the
include statement to keep the "grouping" visible.

Also, if we use `#include "spi.h"` we must add all those include paths
to the makefile and IDE settings (== work, maintainance...). A third
problem would be that we can't have two files with the same name in the
subdirectories, e.g. "per/common.h" and "dev/common.h". Not sure if we
ever need that, but we can leave that door open.

stephenhensley:
Haha, well this got just a little bit out of date. 

Should be fairly trivial, though tedious to resolve. If I have the time in the next day or so I can tackle this. Probably best to get this merged before carrying on with too much else otherwise we'll be chasing our own tails again.

TheSlowGrowth:
@stephenhensley  please take a look again

stephenhensley:
Nice! Thanks for taking the time to clean it up, I've been super busy. 

Should have time to review/merge later today.

stephenhensley:
I should be able to merge this tomorrow, and then we can carry on with everything else.

I caught a build issue while getting the Daisy Field stuff ready that was causing some adc/mux stuff to act really strangely. So I'm going to merge in those changes (mostly just Makefile stuff; should only affect the files that didn't get moved into folders), and then retest with this PR. Might be a solid spot to start keep track of version/releases as well.

stephenhensley:
Not sure how this wasn't didn't prevent the build from working before, but it seems like the lines

```Make
C_SOURCES += $(addsuffix .c, $(MODULE_DIR)/$(C_MODULES))
```

were only adding the `MODULE_DIR` var to the first entry in the list.

Fixed it by adding:

```Make
C_SOURCES += $(addsuffix .c, $(addprefix $(MODULE_DIR)/, $(C_MODULES)))
```

Already checked that the Mux ADC inputs on the Field are working expected, and am going to give a few other things a quick test and then I'll have it merged. 

Sorry again for the inconvenience with the rebasing several times.

PR #226, opened 2020-07-07
Cleaned up QSPI headers

beserge:


stephenhensley:
tested and it still works -- looks good

PR #227, opened 2020-07-09
Daisy Field Update

stephenhensley:
updated daisy field for new hardware and added to makefile. 

Still need to test on the new hardware.


stephenhensley:
This also fixes #228 but we may want to explore other solutions to the problem down the road.

Issue #229, opened 2020-07-11
Release build on Visual Studio not configured correctly


touil:
The Debug config works, but it is useless for the end users.

![image](https://user-images.githubusercontent.com/6236789/87221534-fc992280-c36c-11ea-89a0-0daf6e746aa1.png)

PR #232, opened 2020-07-13
Bug/midi fix

### LINKED ISSUE #218: MidiHandler buffers MIDI messages from UART
I was playing with the [Midi example](https://github.com/electro-smith/DaisyExamples/tree/master/pod/Midi) and hit a frustrating bug. After connecting a MIDI controller over UART (the MIDI jack on the POD), the messages were not available from the `MidiHandler` immediately but needed to get "pushed", i.e. a NoteOn was only getting processed after its NoteOff was sent by the controller.

I think that the delay is caused by the size of DMA buffer set at https://github.com/electro-smith/libDaisy/blob/72827e1b9f3f75031f5320ec5c8ee8a78ba95c78/src/hid_midi.cpp#L34 The only callback that `hid_midi.cpp` sets is triggered when the transfer is complete, and the regular MIDI messages have 2 and 3 bytes only. I think that this means a single message sits in the buffer until full 6 bytes are received.

Adjusting the DMA buffer size from 6 to 1 solves my problem and the MIDI messages are now available to my program immediately. I suspect that it's introducing a risk of losing incoming data and the correct approach would be to rely on the `HAL_UART_RxHalfCpltCallback` callback while setting the data transfer to 2 or 3 bytes but I'm not knowledgeable enough.

stephenhensley:
Fixes #218 

The incoming byte size is now set to 1

In case of overrun, or other errors UART will still disable itself, and will need to be rechecked.

This is built into the MidiHandler for now, but should probably move down into the UART guts.

### DIFF
--- a/libdaisy-Debug.vgdbsettings
+++ b/libdaisy-Debug.vgdbsettings
@@ -23,7 +23,6 @@
     <ProjectFile>libdaisy.vcxproj</ProjectFile>
     <RemoteBuildEnvironment>
       <Records />
-      <EnvironmentSetupFiles />
     </RemoteBuildEnvironment>
     <ParallelJobCount>1</ParallelJobCount>
     <SuppressDirectoryChangeMessages>true</SuppressDirectoryChangeMessages>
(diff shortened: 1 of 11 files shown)


stephenhensley:
As mentioned in commit, this also fixes #229 by providing a working Release VGDB configuration for libdaisy.

PR #240, opened 2020-07-28
Adds interleaved and multi-channel callback support to all boards

beserge:


src/daisy_patch.cpp
```diff
@@ -61,6 +61,14 @@ void DaisyPatch::SetAudioBlockSize(size_t size)
     dsy_audio_set_blocksize(DSY_AUDIO_INTERNAL, block_size_);
     dsy_audio_set_blocksize(DSY_AUDIO_EXTERNAL, block_size_);
 }
+
+void DaisyPatch::StartAudio(dsy_audio_callback cb)
+{
+    dsy_audio_set_callback(DSY_AUDIO_INTERNAL, cb);
+    dsy_audio_start(DSY_AUDIO_INTERNAL);
+    dsy_audio_start(DSY_AUDIO_EXTERNAL);
+}
```
stephenhensley, line 70: did you test this one? It might be a little weird since the callback would be getting called twice (once per codec). 

I don't think the patch should support quad-interleaved. If interleaved should be there at all, it should probably be stereo only, or have two callback arguments.

PR #243, opened 2020-08-06
Skip PDF generation on development builds to save time

TheSlowGrowth:
In CI builds on feature branches, no docs are deployed (DEPLOY_DOCS env var is not set to 1).
Hence, the whole PDF doesn't actually need to be built.
Hence, latex & fonts don't even have to be installed.

This PR fixes that, reducing build time A LOT.


stephenhensley:
Totally makes sense. Looks good

Issue #237, opened 2020-07-27
CD4021 Input Shift Register doesn't like optimization


stephenhensley:
I'm guessing its optimizing out one of the loops or moving some of the GPIO stuff around. Because pushing one button reads in all of them. Setting to `-O1` or `-O0` both work, but `-O2` and above do not.

Should be able to put a dsb or something there to make sure it does everything before moving on.

stephenhensley:
Just a note in case my changes get lost in the mess of PRs I'm about to merge before addressing this directly, but there was no delay between the clock strobe on the back end of the for loop for each data bit. This was either getting optimized out, or it was just going high for so little time that it `wasn't long for the 4021 to catch.

Here's the functional update function:

```C++
// 3000 = 15us max clock/rise/fall time
static const uint32_t kGpioDelayAmt = 1; // in ticks @ 200MHz
void dsy_sr_4021_update(dsy_sr_4021_handle *sr)
{
    uint8_t idx;
    // Strobe CS Pin
    dsy_gpio_write(&sr->clk, 0);
    dsy_gpio_write(&sr->cs, 1);
    dsy_tim_delay_tick(kGpioDelayAmt);
    dsy_gpio_write(&sr->cs, 0);
    // Clock through data.
    for(uint8_t i = 0; i < 8 * sr->num_daisychained; i++)
    {
        // Grab data from each parallel data pin.
        dsy_gpio_write(&sr->clk, 0);
        dsy_tim_delay_tick(kGpioDelayAmt);
        for(uint8_t j = 0; j < sr->num_parallel; j++)
        {
            idx = (8 * sr->num_daisychained - 1) - i;
            idx += (8 * sr->num_daisychained * j);
            sr->states[idx] = dsy_gpio_read(&sr->data[j]);
        }
        dsy_gpio_write(&sr->clk, 1);
        dsy_tim_delay_tick(kGpioDelayAmt);
    }
}
```

stephenhensley:
this was resolved in #244

PR #244, opened 2020-08-12
Pre folder fixes

stephenhensley:
Primary purpose of this PR is to resolve drastic differences between the VisualGDB build and the Make build. Now the release binaries produced by both are the same size, and function the same way.

What made this necessary was very strange behavior when using a Mux for ADCs on the Daisy Field. (channels 7 and 8 worked fine, but they others were partial garbage data most of the time..). Building from VGDB acted as expected. 

There were a few duplicate .o files being added in the libdaisy.a, which seemed like the issue, but resolving that did not fix the problem.

Weirdly, running Make's sort on the `OBJECT` list did resolve the issue (as far as I know, the order of the .o files shouldn't change the behavior, but this should be investigated further still)

the new core of the Make recipe looks like:

```Make
#######################################
# build the application
#######################################

# list of C program objects
OBJECTS = $(addprefix $(BUILD_DIR)/,$(notdir $(C_SOURCES:.c=.o)))
vpath %.c $(sort $(dir $(C_SOURCES)))
# list of CPP program objects
OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(CPP_SOURCES:.cpp=.o)))
vpath %.cpp $(sort $(dir $(CPP_SOURCES)))
# list of ASM program objects
OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(ASM_SOURCES:.s=.o)))
vpath %.s $(sort $(dir $(ASM_SOURCES)))

# Prunes duplicates, and orders lexically (for archive build)
SORTED_OBJECTS = $(sort $(OBJECTS))

$(BUILD_DIR)/%.o: %.c Makefile | $(BUILD_DIR)
	$(CC) $(CFLAGS) $(C_STANDARD) -c $< -o $@ -MD -MP -MF $(BUILD_DIR)/$(notdir $(<:.c=.dep))

$(BUILD_DIR)/%.o: %.cpp Makefile | $(BUILD_DIR)
	$(CXX) $(CPPFLAGS) $(CPP_STANDARD) -c $< -o $@ -MD -MP -MF $(BUILD_DIR)/$(notdir $(<:.cpp=.dep))

$(BUILD_DIR)/%.o: %.s Makefile | $(BUILD_DIR)
	$(AS) -c $(CFLAGS) $< -o $@ -MD -MP -MF $(BUILD_DIR)/$(notdir $(<:.s =.dep))

$(BUILD_DIR)/$(TARGET).a: $(SORTED_OBJECTS) Makefile
	$(AR) -r $@ $(SORTED_OBJECTS)

```

accidentally added the sr_4021 fix for (#237) which will probably cause a conflict in #224, but it'll only be that one file. 

I also added C_STANDARD and CPP_STANDARD flags to the build so that we're not just leaving it to the compiler's default anymore.

I have them set to C++14 and C11 for now. C++17 caused a lot of `register` warnings when compiling HAL.


stephenhensley:
one other thing, I added the entire HAL to the libdaisy.a since there was no clear line of what was or wasn't included.

This does increase the build size/time of the .a file, and we could probably trim certain things out, but I was mostly just trying to make the Make build consistent with the VisualGDB build (which also builds all of the HAL files)

stephenhensley:
Also just documenting the other changes in case we're looking for stuff later:

- Daisy Field fixes for proper knob order, and a few missing functions, etc.
- switched to a different startup file written in C (same as the one used by VisualGDB)

Issue #238, opened 2020-07-27
Excess time spent in AudioCallback causes dropped MIDI bytes


hammondeggs:
Currently, if you spend too much time in your AudioCallback routine, the MIDI UART handler will occasionally drop bytes.

You can reproduce this with the existing Daisy Pod Midi example by having the CPU perform 'busy work' (for example a large FIR or other dummy calculations if you like) during the callback.

For now, I am using a temporary workaround where I simply play a buffer in the audio callback and build this buffer outside any interrupts / DMA transfers as needed. This of course isn't ideal, but it does work, and barely uses any CPU time. 

(For what it's worth I had a look at libdaisy and tried inverting the DMA priority of the UART DMA and the audio DMA but no luck).

stephenhensley:
Good to know! 

I want to look into the UART peripheral and how it interfaces with the DMA a bit more, because it would be ideal to set it up to be able to fill a fairly large FIFO, and still be able to be read with as few as 1 or 2 bytes. 

Right now the hardware is trying to throw an interrupt for every byte received so that it can fill a software FIFO without stalling when there isn't a full buffer.

This will be happening along with some other MIDI-related stuff over the next week or so.

stephenhensley:
I'm going to address this in #250 as well as the running status fix. 

I have a mostly fixed system in place that augments the DMA use to read up to 256 bytes, and uses the IDLE line interrupt to trigger the buffer copy (this interrupt is fired when there is no activity on the bus for the 1 byte-period). 

I'm currently testing, and cleaning up, but it's working very reliably.

stephenhensley:
This was fixed in #250

andrewikenberry:
![](https://github.trello.services/images/mini-trello-icon.png) [Flexible SPI interface (all configurations for Rx and Tx)](https://trello.com/c/tSEWp74r/16-flexible-spi-interface-all-configurations-for-rx-and-tx)

PR #250, opened 2020-09-04
added running status support to MIDI

### LINKED ISSUE #239: MIDI does not appear to support Running Status / NoteOff via NoteOn
I know the MIDI implementation is currently incomplete however I will note a few things anyway in case they are helpful.

Currently, the MIDI implementation appears to require the full 3 byte Status, msg, msg format. E.g. NoteOn, NoteNum, Velocity etc.

To save time, MIDI supports something called "Running Status" where you can send a single NoteOn message followed by pairs of bytes for Note Num / Velocity - allowing you in this example to send multiple note ons (*and possibly note offs as well explained later).

Note as well, that a NoteOn message with a velocity of "`0x00`" is actually processed as a NoteOff (with a velocity of zero). This also means that during a running status message you can have both note ons and note offs.

For example:

`0x90 3C 4A 3D 00 3E 4A`

3 messages are present here in this one group: 
`0x90` - Note On, Channel 1
`0x3C` - Note Num for note on
`0x4A` - Velocity for this note on
`0x3D` - Next Note num (still using `0x90` status)
`0x00` - Velocity is 00 - this should be treated exactly the same as `0x80 0x3D 0x00`
`0x3E`- Next Note num (still using `0x90` status)
`0x4A` - velocity for this note num.

Etc.

stephenhensley:
This should fix #239 

Most of my hardware doesn't seem to use running status, but I was able to find an older oxygen8 that I couldn't reliably use with the MIDI examples before this update, and was actually using the running status feature. 

Testing from a few sources this seems to fix a lot of the 'dropped note' issues. However, I am still finding that if too much time is spent in the audio callback this will cause UART to miss new data coming in. So I'll reopen the issue discussing that problem as we'll probably we want to flesh out thata implementation a bit more to prevent issues like these.

### DIFF
--- a/fix_style.sh
+++ b/fix_style.sh
@@ -2,7 +2,7 @@
 # Uses clang-format-10.0.0
 #TODO:
 # - fix all the hard-coding
-STARTDIR='./src/'
+STARTDIR='./src/**'
 find $STARTDIR -iname '*.h' -type f -exec sed -i 's/\t/    /g' {} +
 find $STARTDIR -iname '*.cpp' -type f -exec sed -i 's/\t/    /g' {} +
 clang-format -i $STARTDIR/*.cpp $STARTDIR/*.c $STARTDIR/*.h
--- a/libdaisy.vcxproj
+++ b/libdaisy.vcxproj
@@ -72,7 +72,7 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|VisualGDB'">
     <ClCompile>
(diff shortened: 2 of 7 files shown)


stephenhensley:
With UART DMA changes, added this also resolves #238

stephenhensley:
Also worth mentioning here that the Rx Callback in UART will copy the FIFO to a new buffer that is accessed by the `PopRx()`, `Flush()`, etc. functions. 

I did test this with direct access to the buffer that the DMA is writing to, and it works. Since the memory/time saving seems negligible (less than 300 bytes of RAM, and less than 200 bytes of flash) I left it buffered to prevent any read while write issues that could happen when processing the buffer. There is, however, a macro to enable/disable this in uart.cpp if desired.

PR #251, opened 2020-09-15
Audio/SAI Redux C++

stephenhensley:
## Summary

Rewrote SAI/Audio with pimpl approach. 

SAI can now be configured with various samplerates (8k, 16k, 32k, 48k, 96k)

Adjusted PLL3P Clock output to be at a high enough rate for all of those samplerates.

Moved the conversion functions (i.e. `s242f()`, etc.) to daisy_core.h instead of daisy.h, and added signed 32 bit conversion functions as well.

SAI can now be used independently without using the `AudioHandle` (if one desires), by passing in buffers, and a User callback. This callback is called directly from the SAI DMA Interrupts, and passes the data as `int32_t` in the bitdepth/format specified by the SAI config.

Previously the Audio handle was actually where nearly all of the audio code was. Now it's essentially a wrapper around the SAI that provides interleaved/non-interleaved callbacks. Not sure if we should leave this separate, or just merge it with the SAI. (Since the actual device driver is now separately configured outside of both of these classes this might just be redundant abstraction).

Given the magnitude of these changes I propose we (finally) begin to version libdaisy. So those concerned with backward compatibility or breaking changes can be aware if/when they occur.

## In Progress

* [x] Multi-codec operation still hasn't been added back in (so the patch examples will only be the first two channels right now.
* [x] DaisySeed interface should allow configuration for setting the samplerate/blocksize since that can only be set at Init time.
* [x] Changing Callbacks while running should still be possible without having to stop/start the DMA as it will be just a function pointer change. The functions just need to be implemented.
* [x] Checks for max block size, etc. in Audio Init

## Impact on Examples/Existing Code

This should not break compatibility with previous code that used the DaisySeed, or other breakout board APIs. However, code that did not use that level of abstraction will require some adjustments to work.

Nearly all examples still build (A few old seed examples that use the `dsy_audio_*` functions need to be adapted)

Multi-Codec support needs to be added for the Patch examples to work correctly again.

DSY_AUDIO_SAMPLE_RATE was removed, and should be deprecated as samplerate can now be configured.

## Other Notes/Issues

The Internal Callback in the AudioHandle class is _very_ long, and adds a lot of potentially unnecessary code size for handling the various possible bit-depths from the SAI. I'm very open to alternatives on handling this.

I hard coded buffers for the Audio handle so that this continues to be easy for beginner users to get audio up and running quickly without them having to configure buffers in the correct RAM section.

Originally, I wanted the Audio Handle to be able to handle the SAI, and Device based on some configuration, but this would require passing in loads of different configs for the various peripherals (sai, gpio, and potentially i2c for certain codecs). So for right now the device should be initialized between the SAI init and the AudioHandle Init.

Protocol/Bitdepth are tied together for now, this could be separated (maybe it should) into separate config options (i.e. bitdepth as-is, and protocol options such as `I2S_STANDARD`, `I2S_MSB_JUSTIFIED`, etc.) I was attempting to provide as much flexibility while still keeping it practical to the expected/common use-cases.

The I2C peripheral uses a global init to configure its software DMA dispatching, but I didn't see this as necessary for the SAI so it was ommited.

Not directly related, but the DMA stream organization is only going to turn into a bigger mess as we add more functionality to libdaisy. This is a better topic for a new github issue, or something, but there is no easy way to tell what streams are for what peripherals, purposes right now without jumping through a bunch of files. Even with only 6 DMA streams set up it's already confusing.


stephenhensley:
@TheSlowGrowth curious on your thoughts on some of this. 😄

andrewikenberry:
[Audio Update (CPP, SampleRates, etc.)](https://trello.com/c/4kIe9a3g/1-audio-update-cpp-samplerates-etc)

PR #253, opened 2020-09-18
Fixed mux'ed ADC readouts being written to the wrong buffer locations

TheSlowGrowth:
Without this commit, I get all sorts of weird readouts from my ADCs. It seems the ADC reads are correct but they end up in seemingly random positions in the buffers. This commit fixes that.

(side note: My setup is this: 6 multiplexers are multiplexing 3 inputs to one adc pin each. All 6 multiplexers are controlled by the same 3 address pins.)

### DIFF
--- a/src/per/adc.cpp
+++ b/src/per/adc.cpp
@@ -133,7 +133,7 @@ struct dsy_adc
     uint16_t mux_index[DSY_ADC_MAX_CHANNELS]; // 0->mux_channels per ADC channel
     // dma buffer ptrs
     uint16_t* dma_buffer;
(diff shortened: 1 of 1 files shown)


stephenhensley:
Fascinating. Interesting that it was such a simple fix.

I can probably get this merged today.

stephenhensley:
Everything builds good, and DaisyField Mux ADCs still test OK.

Looks good!

andrewikenberry:
[Audio Update (CPP, SampleRates, etc.)](https://trello.com/c/4kIe9a3g/1-audio-update-cpp-samplerates-etc)

PR #254, opened 2020-10-07
Enable link time optimizations and improve dead code elimination

TheSlowGrowth:


TheSlowGrowth:
Ouch, this PR turns out to be redundant, the flags are already there.

PR #255, opened 2020-10-10
Improved DrawArc & optimizations

AvAars:
I've been continuing my work on the OLED draw methods, when working on a project I was left unsatisfied with the arc method I ported from the original library. Using approximations gave a jagged result and was a lot (!) slower and uglier than drawing a circle. Especially using animated arcs (peak meters!) were not really up to par.
I have now adjusted the arc method based on the circle method and have changed the circle method to draw a full 360 degree arc, with some optimizations to reduce calculations.
The arc method is even 3 to 4 times faster than the old method in my tests and looks about perfect.

I have also benchmarked some types (int8, int16 and int32, and their unsigned counterparts) and have indeed found that int32 is the fastest (as mentioned in #231) so I worked on that too.
I have currently opted to use the fast versions of (u)int8 and (u)int16 though, to hint at their usable range, but this can easily be changed to normal (unsigned) integers as they all compile down to the same 32 bit integers anyway.

Finally I made the WriteString method take in a const char* to make life a little bit easier (as it shouldn't change any characters anyway).

### DIFF
--- a/src/hid/oled_display.cpp
+++ b/src/hid/oled_display.cpp
@@ -188,7 +188,7 @@ void OledDisplay::Update(void)
     }
 }
 
-void OledDisplay::DrawPixel(uint8_t x, uint8_t y, bool on)
+void OledDisplay::DrawPixel(uint_fast8_t x, uint_fast8_t y, bool on)
 {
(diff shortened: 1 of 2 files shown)


stephenhensley:
Sweet! excited to try to do some more dynamic visualizations and stuff!

PR #256, opened 2020-10-10
cpp standard can now be changed in user projects

TheSlowGrowth:


stephenhensley:
Nice. Definitely useful.

PR #257, opened 2020-10-11
Small fixes (core Makefile, const char*, spelling fix)

jbeda:
Hello! These are some small fixes for things that I noticed as I was playing with my new Daisy Patch.

The linker fix is to allow for `-u _printf_float` so I can get a floating point enabled sprintf.


stephenhensley:
Nice! Thanks for the contribution!

PR #270, opened 2020-11-20
Feature/pcm3060 support

stephenhensley:
Adds driver for TI PCM3060 codec

For now it only has basic config set to work with the DaisySeed default 24-bit I2S handling, and still needs to be tested.

There are some cool features/configuration on the codec that can and should be added (attenuation for each channel, different format specs, oversampling/filter controls, etc.), though I think it will be mergeable once its confirmed functioning as long as there is a config struct added to allow for future configuration without an API change.

Right now the codec drivers are handled independently of the SAI/audio. This is different from how things were set up at the beginning of libdaisy, but works well for most things. 

The only thing that feels strange now is having to set the format/samplerate in multiple places. It would be nice if we could use the SAI config to initialize the Codec, or the Codec config to initialize the SAI. Not sure which of those would work best, and am open to ideas.


stephenhensley:
Also worth mentioning this adds `I2CHandle::ReceiveBlocking` to the I2C driver.

PR #271, opened 2020-11-22
Board api

beserge:
Not done yet, want to get this looked over and have some questions answered before we finalize this and move on to fixing examples.

### DIFF
--- a/src/daisy_field.cpp
+++ b/src/daisy_field.cpp
@@ -151,6 +151,11 @@ void DaisyField::Init()
     dsy_tim_start();
 }
 
+void DaisyField::DelayMs(size_t del)
+{
+    seed.DelayMs(del);
+}
+
 void DaisyField::StartAudio(AudioHandle::InterleavingAudioCallback cb)
 {
     seed.StartAudio(cb);
--- a/src/daisy_field.cpp
+++ b/src/daisy_field.cpp
@@ -89,7 +89,7 @@ void DaisyField::Init()
     for(size_t i = 0; i < SW_LAST; i++)
     {
         dsy_gpio_pin p = seed.GetPin(sw_pin[i]);
-        sw_[i].Init(p, AudioCallbackRate());
+        sw[i].Init(p, AudioCallbackRate());
     }
 
     // ADCs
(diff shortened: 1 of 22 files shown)


beserge:
Needs checking over.

- There are some private objects that might be better off as public objects. I left them private for now as I wasn't quite sure.
- I need to go through and fix the naming conventions (remove _) for the objects that became public.
- Should probably go through and do some more cleaning. For instance the field has some long functions in the header.
- There are some simple functions that are repeated in every board. Should they be moved to the seed, or elsewhere?
- The formatting script needs to be run to please our overlord Travis.

stephenhensley:
> There are some private objects that might be better off as public objects. I left them private for now as I wasn't quite sure.

Yeah, there are definitely a few outliers. What might be the best 'rule' for this is if manipulations to that object outside of accesses to the API would have any effect on the underlying parts of the DaisyBoard itself.

For example, the `callback_rate_` member of the `DaisySeed` class is computed whenever the block size or sample rate are changed. Therefore access to this should be restricted to the API, because changing it on its own would not have the intended effects.

So really, I think pretty much everything in the Daisy Board classes could justifiably be public with a few exceptions, which it seems like you pretty much did.

If there are any specific ones you're not sure of let me know.

> Should probably go through and do some more cleaning. For instance the field has some long functions in the header.

This would be great, definitely worth cleaning up a bit as you go through it. the `ProcessDigitalControls()` implementation can definitely move to the cpp file. 

Related to cleaning, I think all of the `samplerate_`, `blocksize_` and `callback_rate_` members can be removed. I thought I had gotten them out when I did the audio PR a few weeks ago, but it looks like some of them are still floating around. Since the Seed handles the audio including those details, there's no need to maintain multiple sets of that data. The seed does still need the `callback_rate_` member to save from calculating it when it hasn't changed.

> There are some simple functions that are repeated in every board. Should they be moved to the seed, or elsewhere?

If it's repeated in every board, and isn't just a call to a seed function, it should definitely move to the seed, and just have a wrapper in the specific board. Is there a specific function you're talking about?

stephenhensley:
Tested OK w/ DaisyExamples, and oopsy

Issue #265, opened 2020-11-10
Add generic Logging class


stephenhensley:
Mentioned on the [Roadmap Forum post](https://forum.electro-smith.com/t/the-daisy-roadmap/805/8).

Should provide a simple way of starting up a USB or UART connection, and a print function.

This can be added to the DaisySeed so that there are generic `DaisySeed::Print` and `DaisySeed::PrintLine` functions (final names TBD).

AXP:
### High level requirements

1. Should provide out-of-the box logging functionality over USB
2. Should initialize the required hardware
3. Should support disabling at compile time with a result of zero overhead (API calls should not produce any instructions)

### API requirements

1. API proposal:
```
void Print(const char* format, ...);     // print formatted string without appending end-of-line termination
void PrintLine(const char* format, ...); // print formatted string automatically appending end-of-line termination
void Start(UsbHandle::UsbPeriph device, bool wait_for_pc); // start the logging session. Device is either UsbHandle::FS_INTERNAL or UsbHandle::FS_EXTERNAL. The wait_for_pc flag blocks the execution until a key is pressed in the log terminal.
```
2. API calls should not lead to catastrophic failures if called (or not called) in wrong order/way. The application should still proceed with its core functionality.
3. There should be a way to configure end-of-line sequence ("\n", "\n\r", "\r\n", etc.). Supposedly with an internal #define 
4. If Print or PrintLine functions are executed before the Start() is invoked, the data should be buffered until the interface is initialized (useful for logging from global object constructors)
5. No dynamic memory allocation

### Open questions

1. All methods could be made static, should we make it a requirement?
2. It can be implemented as a separate class with static members only, so invocation in a random place might look like:
`daisy::Logger::Print(...);`
and that would work even if the Seed class is not instantiated in the current scope.
Otherwise (or in addition to the above) it can be pulled into the Seed class, so it becomes simply:
`seed.Print(...);`
3. PrintLine may unconditionally insert the end-of-line characters, or may analyze the string and only insert one if it's not there. It's pretty cheap since after formatting the string the position of the last character is already known.
4. If the Print* call is made while the USB is busy, it will block until it is free. Another approach would be to buffer data and retry on the next Print* call. But that could leave the very last message stuck forever in the buffer. Or we should use periodic tick to retry the transfer.
5. Should external USB port work? On my Pod it doesn't seem to out of the box. Should we support an option to print to both USB ports simultaneously? How about separately, with two instances of Logger?
[...]

PR #272, opened 2020-11-24
Usb logger

AXP:
Implemented generic logging class to resolve https://github.com/electro-smith/libDaisy/issues/265
All requirements seem to be satisfied.
Note that logging is enabled by default, but would only increase the executable size if the application starts the logging session.

Example application PR in DaisyExamples: https://github.com/electro-smith/DaisyExamples/pull/119


stephenhensley:
Woah, huge PR awesome!

I can give it a full review/test it with the example you made as well, but it looks great on first glance.

A few immediate question do come to mind, though (some of which are performance/code size related):

1. Can we have the logger be a part of the `DaisySeed` via composition instead of inheritance? 

Mostly just because that would keep it consistent with the rest of the library, and reduces some complexity on the legibility end. Are there any behavioral differences other than having to manually add the `DaisySeed::Print(const char* format, . . .) { logger.Print(format); }`, etc.  functions?

And I guess a related question would be is there any performance difference involved in either approach? 

---

2. I love the `FLT_AVR`, `FLT_FMT` macros as an alternative to linking the newlib printf with floats feature. We do get a few requests about that fairly often.  Though, I'd be curious what the size difference of just adding that native support instead of having to use work arounds like this. (Especially since the code size only increases if you're using the logging).

Related to that, I see the `FLT_VAR` macro uses `abs`, and `pow` but math.h is not included. Also, it would probably be beneficial to use the float versions of those math functions.

---

3. I'm interested in your thoughts on the future of the USBHandle, (or other USB support) for the Daisy platform.

The next most requested USB feature would Audio/MIDI. And I've been debating whether we try to use the usb handle as a global configurator for the USB of the whole system, or whether we have different classes for different USB classes. The other thing to consider is how we may want to handle/support composite classes (for example, Logging and MIDI).

This doesn't directly apply to this update, but I think its worth talking about now while the usb support is still quite minimal.

---

4. I don't see CI updating for your build. I can run the relevant tests on my end while reviewing, but since you've mentioned a few other contributions coming up it might be worth activating it. Though we will be migrating to Github Actions shortly. So that's up to you.

I bring it up specifically because your Doxygen comment syntax is a bit different then the rest of the library, and I'm not sure if the build of doxygen we use will catch the `// ` comments or not. 

And if you want to do the actual style check yourself you can run `./ci/run-clang-format.py -r src` from the root directory. the Fix style script will basically do an inline replace.

Generally, the style for the documentation is:

```C++

/** This is a single line above a function */
void SomeFunction();

/** And another that has arguments
 *   \param val something to pass in
 */
void AnotherFunction(var_t val);

/** And for comments in the same line */
enum Letters
{
    A, /**< The first letter of the alphabet */
    B, /**< The second letter of the alphabet */
};

```

AXP:
> Woah, huge PR awesome!
> 
> I can give it a full review/test it with the example you made as well, but it looks great on first glance.

Thank you for the kind words and great feedback! That is my first open-source contribution ever, first time using github and one of the first times programming for ARM.

> A few immediate question do come to mind, though (some of which are performance/code size related):
> 
> 1. Can we have the logger be a part of the `DaisySeed` via composition instead of inheritance?

I too dislike how it sticks out. I'll see what can be done (see below)

> And I guess a related question would be is there any performance difference involved in either approach?

I believe even a clean wrapping as you suggest can be made efficient if the wrapper is in the header and thus visible for the compiler. Most probably it will be optimized away. But that would require adding another pair of Print* functions with variadic argument, otherwise it's impossible to wrap them. Similar to how vprintf family of functions do it.
But there's modern way to do it via variadic templates. It's very effective but I need to check first that it doesn't blow up the code size.

> 
> 1. I love the `FLT_VAR`, `FLT_FMT` macros as an alternative to linking the newlib printf with floats feature. We do get a few requests about that fairly often.  Though, I'd be curious what the size difference of just adding that native support instead of having to use work arounds like this. (Especially since the code size only increases if you're using the logging).

This workaround would scale with the number of different variables as it injects some code around each printout call (about 50 bytes I think). The CRT support for floating point formatting on the other hand increases the executable size by about 8kB. You can print a lot before they catch up! :) 

>
> Related to that, I see the `FLT_VAR` macro uses `abs`, and `pow` but math.h is not included. Also, it would probably be beneficial to use the float versions of those math functions.

It would be a disaster if it would be executed like that. It's optimized down to a single VABS instruction (the result is reused 3 times) and a multiplication by a constant 10^x, because x is known at compile time. There is still quite a lot of overhead for converting between float and integer, and there's a also a redundant comparison with 0 to pick the '-' character since unfortunately VABS instruction doesn't modify any flags.. But overall it's pretty decent as it is.

I've added extensive profiling to the Logger example in the DaisyExamples repo. I think it is already visible in the PR. Here is the exempt:
```
LOGGER_EXTERNAL - Print     (  empty string): 0.404 us/call
LOGGER_EXTERNAL - Print     (   text string): 0.516 us/call + 0.010 us/char
LOGGER_EXTERNAL - Print     ( integer value): 1.921 us/call
LOGGER_EXTERNAL - Print     (unsigned value): 1.905 us/call
LOGGER_EXTERNAL - Print     ( float.3 value): 4.675 us/call   <-- this is printf("%.03f")
LOGGER_EXTERNAL - Print     (FLT_VAR1 value): 4.194 us/call
LOGGER_EXTERNAL - Print     (FLT_VAR2 value): 4.382 us/call
LOGGER_EXTERNAL - Print     (FLT_VAR3 value): 4.520 us/call   <-- this is the FLT_VAR3 macro
LOGGER_EXTERNAL - Print     (FLT_VAR4 value): 4.644 us/call
```
So the above macros are quite on par with the CRT implementation

> 
> 1. I'm interested in your thoughts on the future of the USBHandle, (or other USB support) for the Daisy platform.
> 
> The next most requested USB feature would Audio/MIDI. And I've been debating whether we try to use the usb handle as a global configurator for the USB of the whole system, or whether we have different classes for different USB classes. The other thing to consider is how we may want to handle/support composite classes (for example, Logging and MIDI).
> 
> This doesn't directly apply to this update, but I think its worth talking about now while the usb support is still quite minimal.

I do have some thoughts, but I need some more time exploring everything before I'm comfortable voicing architectural opinions.

> 
> 1. I don't see CI updating for your build. I can run the relevant tests on my end while reviewing, but since you've mentioned a few other contributions coming up it might be worth activating it. Though we will be migrating to Github Actions shortly. So that's up to you.
 
Sure, thanks for the pointer. I will look into it.

> I bring it up specifically because your Doxygen comment syntax is a bit different then the rest of the library, and I'm not sure if the build of doxygen we use will catch the `// ` comments or not.

I'm not sure either. I've tried generating the docs yesterday, and running the formatting script you point out below, but it didn't work. I'm on Windows and miss some/most of the required packages. I will spend some time on it and CI in the coming days.

> 
> And if you want to do the actual style check yourself you can run `./ci/run-clang-format.py -r src` from the root directory. the Fix style script will basically do an inline replace.
> 
> Generally, the style for the documentation is:

I was subconsciously adapting to what I was seeing around, but apparently habits die hard :)

stephenhensley:
Wow, well we're glad to have your contribution! And welcome to Github and ARM programming!

So it looks like unless there are virtual functions the function calls to the Logger base class will be linked statically, without any v-table overhead or anything. 

I'd say its still worth exploring what it would look like brought in as composition, and/or via variadic template (which looks like it would be trivial, but as you mentioned the code size could be affected pretty drastically) since I am hesitant to start introducing inheritance unless there is a really good reason to do so. Not because its necessarily bad, but because it may encourage its use in places where it might be a detriment. 

That said, depending on what the wrapper code will have to look like, and whether the bloat from the variadic templates are usable this may be a good exception allowing for inheritance.

---

The performance you marked seems great. I did see the example, but I think I skimmed over the output. Cool to see it on par with the CRT implementation, and the compiler optimizations you mentioned make sense. 

---

And no worries on the CI/Style stuff its easy to stick with what you're used to without even realizing it haha. (I only recently started using doxygen, and prior to that almost exclusively used single line comments for everything). 

For the most part everything else looks conformant to our standards it was just the type of comment that stood out.

As far as CI goes, I'm pretty sure all you need to do is go to travis-ci.com and activate the repos you want to use it, then it'll just run checks every time you push.

we use clang-format-10 for the style which I'm pretty sure I got [here in the pre-built binary section](https://releases.llvm.org/download.html) for windows.

And I could have sworn I had doxygen installed, but am not finding it at the moment. I will be making a wiki/tutorial page very soon that covers installing some of these supplemental tools for those interested.

AXP:
I've explored the ways to get rid of inheritance. I'm detailing 4 different ones below. All code snippets are exempts from DaisySeed class declaration. Only Print is shown, PrintLine is similar of course. 
All 4 have similar performance, I couldn't detect any difference. Surprisingly, option 1 produces the smallest and option 4 the largest executable size.
The only benefit of option 4 is uniform look with other handles.
Options 2, 3 and 4 require adding PrintV function to the logger class that takes va_list as a second argument.

What would you prefer? I'd pick either 1 or 2 as I don't see a point in adding a member variable handle.


A common shorthand for class specialization:
`using Log = Logger<LOGGER_INTERNAL>;`

1. static variadic template function
```
template <typename... VA>
static void Print(const char* format, VA... va)
{
    Log::Print(format, va...);
}
```
2. static variadic function
```
static void Print(const char* format, ...)
{
    va_list va;
    va_start(va, format);
    Log::PrintV(format, va);
    va_end(va);
}
```
3. static variadic function with static handle
```
static Log log_handle;
static void Print(const char* format, ...)
{
    va_list va;
    va_start(va, format);
    log_handle.PrintV(format, va);
    va_end(va);
}
```
4. Non-static function with non-static handle
```
Log log_handle;
void Print(const char* format, ...)
{
    va_list va;
    va_start(va, format);
    log_handle.PrintV(format, va);
    va_end(va);
}
```

AXP:
So far I've implemented option 1.
Also made doxygen happy, now I need to please travis...
more specifically clang-format.
the things it proposes look dumb. I need to find a workaround.

AXP:
Seem to be ready for merge

stephenhensley:
Awesome!

Was away from computers for the past few days, but this is exciting!

I was surprised as well to see the code size for the variadic template be the smallest! But that's good news, after reading up on them it sounds like the best option as it will also do some compile time checking for format types, etc. and doesn't have the runtime overhead of generating/processing the `va_list`.

I should have time to test today or tomorrow.

Also, not directly related, but you may be happy to hear (especially after any frustration of the style cleanup). 

I will be documenting installation/local testing the docs/style stuff locally without having to do a commit-wait-check every time this week (and I actually have a pipeline i'm going to be testing for just inplace fixing of the style through CI. so it'll automatically fix style issues on pushes/PRs to master.

stephenhensley:
Oh, I see you caught an issue. I was going to try to squeeze in some time near the end of the day to test this today. So good timing!

Looks like the DaisyExamples PR is also up to date. So should be pretty easy to test. Provided it works as expected I should be able to get it merged today as well! 😄 

We may want to add the wrapper functions to the other breakout boards as well, but #271 touches all of those, and it'll still be possible to do `hardware.seed.Print("Foobar\n");` from any of the breakout boards as it is now. So probably best to wait until both PRs are merged to extend.

AXP:
Indeed, it's impossible to catch it in a busy unit test scenario, but I've stumbled into it as soon as I started using it in real apps. I figured I'd rather submit it now before you started testing it rather than doing it post-merge.
The DaisyExamples PR is ready too, but there's one caveat. Currently I don't see that repo to  trigger the Travis checks (there's no script there), so the code style is probably not compliant. Not sure if that was intended.

Automatic code style enforcement is not necessarily a good idea. There's at least one place in my code where the current style checker affected the program behaviour and I had make an exclusion waiver. It's the float formatting macros. It's treating the unquoted string as a statement and "fixes" the whitespace, breaking the functionality of the macro.

A guide you mention would be really cool. One problem I have with running the code style checks locally is that it processes files in place and thus triggers git to think all of them are modified. It's afterwards inconvenient to manually revert unaffected files.


Regarding exporting the functions to breakout boards, I think that either all "seed" services  need to be exported (in the simplest form by inheritance), or be accessible via the "seed" member (as you point out is currently the case). Maybe wrapping a few of the most used ones can be done, but we have to be careful not to bloat things.
In my apps, I usually do something like:
```
static DaisyPod  pod;
static DaisySeed& seed = pod.seed;
```
so I have handy access to both.
Since I know how the logger implemented, I also do things like
`static daisy::Logger<> log;`
and use that anywhere without `seed` even being in the scope, e.g. inside DaisySP for debugging.

stephenhensley:
>Currently I don't see that repo to trigger the Travis checks (there's no script there), so the code style is probably not compliant. Not sure if that was intended.

I _just_ set up the CI on DaisyExamples the other day, and it uses GithubActions instead (libdaisy, and DaisySP will likely follow suit very soon). This also allowed me to separate the style check from the rest of the build. So it will still give us a partial green light even if it fails style.

>Automatic code style enforcement is not necessarily a good idea. There's at least one place in my code where the current style checker affected the program behaviour and I had make an exclusion waiver. 

This is a really good point. Before adding it I had considered the possible issues it could cause, and the only time I'd seen it do anything really bad was related to some longer macros I had, but the end result was still functional..

I'll take a look at what it'll look like before I do any merging, but it's trivial to remove the automatic styling, and move that back to a script that can be run.

The only issue will be trying to get it to 'pass' the style check despite whatever formatting issue it thinks it's fixing. (clang-format has a ton of options, so there's a good chance we can fix it that way).

>A guide you mention would be really cool. One problem I have with running the code style checks locally is that it processes files in place and thus triggers git to think all of them are modified. It's afterwards inconvenient to manually revert unaffected files.

Indeed! I'm hoping to get to that early next week or over the weekend if I have some spare time.

I did a fair amount of documentation on my [CI test repo's README](https://github.com/stephenhensley/DaisyActionTest) if you want to look that over in the meantime.

In DaisyExamples, I added a `local_style_check.sh` that runs the equivalent style check command (but without the in-place fixing that the libdaisy fix_style.sh does). That said, the `./ci/run-clang-format.py -r src` command I mentioned a while ago shouldn't do any modifications to the files if run in the libdaisy repo. Maybe it's doing something to the line endings or something?

>I think that either all "seed" services need to be exported (in the simplest form by inheritance), or be accessible via the "seed" member (as you point out is currently the case).

It's definitely a balancing act to make it as easy as possible for beginners, and not overly bloated, or complicated. I'm still looking into the ramifications of non-virtual inheritance for things like breakout boards, but I'm hesitant to introduce it in any form as it would set a precedent for other uses further down the road that may/may not have negative side effects on performance, etc.

Personally, I don't mind using it strictly via composition (i.e. `hw.seed.Print()`), but this is another layer of indirection for beginners (and inheritance does fix that, but then finding 'where Print comes from' is a bit more challenging). So for the very common API I like the idea of just having wrappers extended down to the breakout boards (i.e. `StartAudio()`, etc.)

Using the static version is also pretty user friendly (once its declared). As that's pretty similar to how a lot of the arduino libraries work. (You just say 'Serial.begin()` without having to declare the mysterious Serial object.

----

All that said, doing a quick read-through everything looks pretty good. I'll be able to look into the potential style issue on the DaisyExamples, and hopefully get everything merged tomorrow morning.

AXP:
Thanks for sharing you thoughts, I'll have a look tomorrow.
Small note: the styling mishap is in the libDaisy code, not DaisyExamples.
here it is: (taken from logger.h)
```
// clang-format off
#define FLT_FMT(_n) STRINGIZE(PPCAT(PPCAT(%c%d.%0, _n), d))
// clang-format on
```

stephenhensley:
Oh! Must have misunderstood. Good to know; thanks! 

Didn't know you could enable/disabled clang-format like that; pretty cool.

stephenhensley:
Alright!

Having such a thorough example made testing this a breeze! 

Everything seems to work as expected, and the example output matched what was expected based on the print at the bottom of the example source.

I figure I'll just make notes about both the libdaisy addition and the example here rather than split the review up.

The Visual Studio project needed a little bit of massaging to build correctly, which actually illuminated some really minor issues that I wouldn't have otherwise noticed (a 2 and 3 really only have an effect on the VS build, but are simple enough to resolve that we might as well.

1. It looks like `usbd_def.h` is being included only for `USBD_OK`, (which the existing hid/usb.h is to blame for because it implicitly returns it as a `uint8_t`. (This is the first time that the Middleware/USB stuff becomes necessary to include in the user project).  A solution would be to just add a `UsbResult` type/enum to hid/usb and utilize that. Not a huge deal either way, just makes configuring builds outside of the `core/Makefile` a bit more involved. 
    - Related to this, most of the internal bits of libdaisy and the ST HAL have their success results coded to 0, with all error codes as non-zero. It looks like your `bool Transmit` functions flip this (though, I agree that in the context of a bool, `false` would imply failure.)
2. I only saw this in VS, but I was getting an undeclared error for `va_list` which wasn't happening with the Makefile build, but can be resolved by adding `#include <cstdarg>` to the logger.h
3. Same as above, but for sprintf in the DaisyExamples/seed/Logger/ file, again resolved with `#include <cstdio>`  I've seen the MSBuild version be a little needier in terms of standard library stuff, but I'm still not sure why.

Not an issue, but I may be confused about the destination usage in the example, but it shows profiling the semihosting/external ports even if those aren't configured for use. 

Other than build changes everything looks good and works for the internal. 

I still need to test the external/semihosting destinations in a separate, but don't anticipate finding anything else of note.

---

Re. style stuff. I'm going to update DaisyExamples to not do in-place style fixing before I merge the example, and I'll add a similar `fix_style.sh` as we have in libdaisy. That way it can be used/tested prior to pushing to avoid any unforeseen issues.

At first glance it looks like the style check will fail for a few things (mostly long function argument lists on a single line, and multiple statements on a single line, etc.). So it looks like it'll initially fail style, but you can probably just checkout/run the fix style script once I've added them to master. I can post the git commands for doing that when it's ready to go.

To style-check just your example on the current commit I ran the following:
```
git checkout master util/run-clang-format.py
./utils/run-clang-format.py -r seed/Logger
```

I didn't test it, but you could probably run the fix by doing:
```
clang-format -i seed/Logger/Logger.cpp
```

stephenhensley:
Oh, one other thought that just occured, though it doesn't look like it has an impact, but will we still be able to implement a separate Serial Receive class to handle receiving messages from the host while using the Logger class? Just trying to think forward to other use-cases as well.

AXP:
Thanks a lot for testing!
I'll check and fix the includes, thanks for pointing out.
The uint8_t/bool thing is an interesting debate. I think if we don't enforce uint8_t as return type everywhere (which BTW could be inefficient on 32-bit architectures if the compiler doesn't extend it automatically to 32 bits), then it would often be tempting to use bool as return value. And it makes sense for `true` to be positive result. With uint's you never know how uniform all the components are, so I always try to use actual define/enum value rather than assuming 0 should be treated as OK. And I also believe that if any module returns integer value, a header with define/enums should be automatically included. So hid/usb.h should indeed do that.

Regarding the logging destination usage in the example, I should have made it clearer.
The ideas was to a) verify that using uninitialized destination doesn't cause catastrophic failures and b) profile performance of different things.
The performance stuff is probably not very apparent, the idea is:
1) The uninitialized port (LOGGER_EXTERNAL) would profile my wrapping and formatting code excluding the actual UART transmission
2) the stdout port (LOGGER_SEMIHOST) would profile same wrapping but with additional write() call

I'm not too optimistic with the external USB and semihosting. I couldn't get the former to work at all on my Pod, it's just not visible from the PC side. Maybe I'm missing something. Same with semihosting, I don't have an ST-compatible debugger at hand, so i cannot check it myself.

AXP:
> Oh, one other thought that just occured, though it doesn't look like it has an impact, but will we still be able to implement a separate Serial Receive class to handle receiving messages from the host while using the Logger class? Just trying to think forward to other use-cases as well.

It shouldn't be a problem. Matter of fact, I did is as part of Logger at first, but then decided to drop it. I believe it was totally orthogonal to transmitting given the current USB implementation. I may do such standalone Receiver class later on myself, if nothing else comes up.

stephenhensley:
Great! Look forward to what may be the last few changes then! Very excited to be able to start using this. 

Re. the uint8_t/bool issue. I 100% agree about the `uint8_t` inefficiency, and favor the enum/Result type (as seen in most recent additions/modifications to the lib like SAI, I2C, Audio, etc.) using the 8-bit type is just an old habit from working on smaller micros that I still catch myself doing.

That makes total sense in regard to using the other destinations for profiling. 

I think I have some stuff that I can turn semi hosting on fairly easy to test that, and I'll look into the external port not showing up. I do have older firmware (used for testing the pods) that uses that port for Serial. So it definitely works, it could be some interaction with the the internal port. I'll look into that a bit, but we can fix that independently if need be.

And that's what I figured about receiving, but I wanted to mention it now. Before we wrap this up and realize we didn't think of something.

AXP:
That was the last planned update, also merged in latest master to make sure it's still compatible.

stephenhensley:
Awesome! Looks good!

I'll give it one more quick test in an hour or two, and then get it merged (We can fix the retval stuff in a PR. There are a few other earlier peripherals that may also need that cleaned up a bit).

stephenhensley:
Alright. Retested the example and gave the Visual Studio example a go as well.

With some modification to the vcxproj the VS studio build works, and the module itself seems solid. I'm signing off!

---

The `usbd_def.h` still requires including a lot of stuff that weren't required (listing here for self-documentation since this should be resolvable in the upcoming patch to UsbHandle that adds proper Status enum):

* libdaisy/Middlewares/ST/STM32__USB_Device_Library/Core/Inc
* libdaisy/Drivers/CMSIS/Device/ST/STM32H7xx/Include
* libdaisy/Drivers/CMSIS/Include
* libdaisy/Drivers/STM32H7xx_HAL_Driver/Inc

Until this patch, most VisualStudio/VGDB projects will need those things added to their Include Directories.

PR #276, opened 2020-12-04
Better Mux handling in ADC

stephenhensley:
There was an issue that never entirely went away (much noticeably worse with DEBUG set in the Makefiles, and Debug configuration in Visual Studio).

Data from one input on a multiplexor would get read into the wrong channel since the conversions could happen faster than the callback would. In non-debug builds this wasn't really noticeable, but there was slight flicker in LEDs being controlled directly by ADC channels.

This basically just adds a stop/start condition with a reinitialization of the DMA.

Longterm, this should get rewrittten similarly to how the I2C driver handles this sort of thing, and probably use a timer as a steady timebase, but this works for now.

----

There should be no functional difference for the boards that don't use a multiplexor as the DMA will still be configured to be fully circular and no callbacks are necessary.

### DIFF
--- a/src/per/adc.cpp
+++ b/src/per/adc.cpp
@@ -363,15 +363,9 @@ void AdcHandle::Init(AdcChannelConfig* cfg,
 
 void AdcHandle::Start()
 {
-    size_t rx_size;
     HAL_ADCEx_Calibration_Start(
         &adc.hadc1, ADC_CALIB_OFFSET_LINEARITY, ADC_SINGLE_ENDED);
-
-    // When using the mux we use the HalfCplt callback to fill data, and the pins change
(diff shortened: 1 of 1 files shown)


stephenhensley:
Okay. A few more changes, all within the ADC module as well:

* Moved mux_used setting to earlier in the Init so that that can control some of the ADC config
* Added internal static function for initializing DMA so that it can be done repeatedly when the mux is used.
* Added necessary changes to the ADC Init Config  for the pseudo-circular behavior when using Mux.

This has been tested on the DaisyField and DaisyPatch to ensure that changes to the Multiplexor behavior did not have any adverse effects on the normal operation in circular buffer mode.

There are a lot of other things I want to add to ADC, and it should get updated to look more like Audio/I2C, etc. but that is for a different update.

PR #285, opened 2020-12-15
UsbHandle retvals, fixes need for usbd_def inclusion in user facing files.

stephenhensley:
Added `UsbHandle::Result` enum class that matches I2C, Audio, etc. 

Replaced internal usage, updated headers, and fixed in Logger implementation as well. That allows us to remove the usbd_def.h include within logger_impl.h, _greatly_ reducing complexity of VS builds, etc. by limiting necessary include directories in user projects.


stephenhensley:
Minimal PR -- rebuilt everything, checked that VS projects can be simplified as intended (only affects  VS projects since logger was added), and retested Logger example to make sure everything is still working as expected.

PR #286, opened 2020-12-17
System/Clock configuration

stephenhensley:
Updating the system stuff to cpp, with a configuration struct similar to Audio, I2C, etc.

So far this only consists of:

* CPU freq
* ICache enable/disable
* DCache enable/disable

Ak4556 needed to get the cpp treatment as well, and I made the singular function static within a class so it can be called as:
`Ak4556::Init(pin)` for now since there is no data, or anything else in the class. Debatably it could be a function without a class, but I like the consistency with the other modules.

Also related the following functions were moved into the new System class but also made static so they can be called whereever they were being called before without having too have access to the object:

* `dsy_system_getnow()` is now `daisy::System::GetNow()`
* `dsy_system_delay()` is now `daisy::System::Delay()`

I'll be testing the faster clock speed and going over everything again before this gets merged. But I figured I'd open this early if anyone wanted to look it over.


stephenhensley:
this will probably fail style because I had stuff open with visual studio when working on it. so its probably filled with tabs.

stephenhensley:
Boosted clock mode checks out so far. Easily able to convert existing projects to boosted 480MHz by updating Init to:

```cpp
daisy::DaisyPod hw;

int main(void)
{
    hw.Init(true); // boolean setter for boost mode
}
```

This works for DaisySeed and all supported breakout boards.

---

Before this can get wrapped up I still need to do the following:

* [ ] Update the `per/tim` source to cpp, and have it check the appropriate clock to determine the tick rate. 
    * this needs to be updated to cpp at some point anyway, and it needs access to the new clock frequency getters in System.
* [ ]  Test SDRAM, and compare performance to ensure new clock rate is functioning as expected (Bumped up to max 100MHz from 87MHz). **Test this at 400MHz and 480MHz sysclk**
* [ ] Test QSPI and make sure it is still functional with sysclk at 480MHz (double check its source clock and make sure it is still below the max 133MHz supported by the chip).

stephenhensley:
On top of Tim -- the required changes keep piling up.
SR4021 requires tim tick-delay. We can maybe work around this. Otherwise, it also needs to get updated to cpp.

stephenhensley:
Alright in the interest of having a good timer source for measuring stuff (sub millisecond) I'm going to add the TIM2 stuff (previously dsy_tim) to the new `System` class, and make several of the timing getter/delay functions static.

The Base  `Delay` and `GetNow` will return the SysTick equivalent, while the other more explicit forms will use TIM2.

This will allow people to adapt their:

`dsy_system*` and `dsy_tim*` code to `System::*` without having to use the seed.timer object, or anything more complicated. This also simplifies use in the context of internal modules that needed really small delays of just a few ticks.

This can also lead to just removing the SysTick at some point if we don't need it (it is 1000 IRQs a second that aren't doing anything more than calling a `__weak` callback, and updating the 'tick'.  At some point I'll be adding the ability to attach a callback to a periodic TIM update which would cover that use case as well.

stephenhensley:
Just incase we're ever tracking down why the Nanosecond getter/delay have been removed from the timer stuff, its because as soon as prescaler, is changed, or the sysfreq is not an even value (the case when set to 480MHz), nanoseconds will always be fractional, (1.2ns per tick instead of 2). Instead conversion can be done on the final values of tick-based calcs when nano-second units are necessary in the user application).

stephenhensley:
Check QSPI and the signal looks a little more jittery at the new frequency but still is within  the max freq of the chip, and it still passes a 4MB write/read test.

Currently the QSPI Clk signal will be 100MHz when the sysclk is 400MHz, and 120MHz when the sysclk is 480MHz.
Max for the chip is 133MHz.

This will get setup to a different clock source in January when we flesh out the QSPI support more that can fix it at 130 (or close), or if performance of DTR is better we can have it use that instead at the lower maximum rate for that.

stephenhensley:
I2C timing values did have to get adjusted, and weirdly the measured value when PCLK is at 120MHz is actually a few tens of kHz slower when running fastmode+. The issue is the rise-time proportionate to the signal timing, and testing with 4k7 pull ups I'm still unable to get it up to the 1MHz point, but that's nothing new.

PR #288, opened 2021-01-05
Fix documentation for MidiEvent::AsControlChange

kisielk:


stephenhensley:
good catch! Thanks for the fix! 😄

brianpowalski360:
Hi it's brian powalski

PR #292, opened 2021-01-08
Migrating CI to GIthub Actions

stephenhensley:
This will actually be nice because we can have clear visualization of the three tests (style, build, and documentation), and who doesn't love being able to add more badges to things.


stephenhensley:
note to self: It would be very handy to get [nektos's act](https://github.com/nektos/act) running on my local machine to test actions locally

stephenhensley:
Okay, seems like all three tests are functioning.

I do want to change the docs build so that it only actually deploys the docs on push to master, though it should still make sure they can build on any PR to master.

PR #296, opened 2021-01-21
Built-in DAC - CPP and DMA.

stephenhensley:
updated Dac peripheral to cpp with pimpl. 

working in polling mode need to add dma mode (Add internal TIM5 usage for triggering the DAC, add DMA stream channel/Init, and provide some way of choosing a samplerate).

all old code is still in place, but may not function properly (I already overwrote parts of the MspInit). So that also needs to get cleaned up quite a bit.


stephenhensley:
More progress - DMA is working (tested at 48kHz on channel 1). Still need to test channel 2 and the dual output configuration, but its pretty nearly good to go.

I do want to see about moving the actual enum class definitions for the Config outside of the Config struct. i.e. something like:

```C++
class DacHandle
{
    enum class Mode
    {
        POLLING,
        DMA,
    }
    struct Config
    {
        Mode mode;
    };
};
```

That will simply the syntax to something a bit more familiar, especially given that things like `Channel` get used outside of the class/configuration. 

There's really no added legibility to `DacHandle::Config::Mode::DMA` than there is to `DacHandle::Mode::DMA` and the latter is just as clear while being shorter.

stephenhensley:
I've made the changes mentioned above about pulling the `enum class`es out of the config struct, and I think everything reads a lot better, from initialization to setting some of the settings in practice. 

Since everything will still be scoped to the class itself I don't anticipate this causing any issues with the other modules that use this type of structure, and it would only simplify some of the setters. If we do stick with this, I will want to go back through and make the other modules conform for consistency sake.. which may require user code to change..

PR #308, opened 2021-02-22
add CMSIS DSP include paths, remove duplicate defines

AXP:
1. Cleaned up USE_HAL_DRIVER and STM32H750xx that were defined twice
2. Added symbols from libdaisy/Makefile that better define the core configuration:
CORE_CM7, STM32H750IB, ARM_MATH_CM7, USE_FULL_LL_DRIVER
3. Forced inclusion of stm32h7xx.h file, which defines core configuration to make sure all files being compiled are aware of core extensions. In future this simplifies adding CMSIS DSP sources to examples and user projects.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -53,6 +53,10 @@ CPP_SOURCES ?=
 C_SOURCES += \
 $(SYSTEM_FILES_DIR)/startup_stm32h750xx.c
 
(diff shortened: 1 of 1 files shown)


stephenhensley:
Solid. Confirmed this doesn't seem to introduce any issues. Seems good to go. Thanks!

PR #309, opened 2021-02-25
Added driver for LCD HD44780

StaffanMelin:
Added driver for LCD HD44780.

From conversation with shenley as per 19th of Jan 2021.

### DIFF
--- a/src/dev/lcd_hd44780.h
+++ b/src/dev/lcd_hd44780.h
@@ -1,0 +1,46 @@
+#pragma once
+#ifndef DSY_LCD_HD44780_H
+#define DSY_LCD_HD44780_H
+
+#include "daisy_core.h"
+#include "per/gpio.h"
+
+
+class LcdHD44780
+{
+
+public:
+
+    LcdHD44780() {}
+    ~LcdHD44780() {}
+
+    struct Config
+    {
+        bool cursor_on;
+        bool cursor_blink;
+        dsy_gpio_pin rs, en, d4, d5, d6, d7;
+    };
+
+	void Init(const Config&);
+	void Print(const char *);
+	void PrintInt(int);
+	void SetCursor(uint8_t, uint8_t);
+	void Clear();
+
+private:
+
+	bool cursor_on;
+	bool cursor_blink;
+	dsy_gpio lcd_pin_rs;
+	dsy_gpio lcd_pin_en;
+	dsy_gpio lcd_data_pin[4]; // D4-D7
+
+	void WriteData(uint8_t);
+	void WriteCommand(uint8_t);
+	void Write(uint8_t, uint8_t);
+
+};
+
+
+
+#endif
--- a/src/dev/lcd_hd44780.h
+++ b/src/dev/lcd_hd44780.h
@@ -8,39 +8,34 @@
 
 class LcdHD44780
 {
-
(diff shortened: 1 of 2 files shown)


src/dev/lcd_hd44780.h
```diff
@@ -0,0 +1,41 @@
...
+        bool         cursor_on;
+        bool         cursor_blink;
+        dsy_gpio_pin rs, en, d4, d5, d6, d7;
+    };
+
+    void Init(const Config &);
+    void Print(const char *);
+    void PrintInt(int);
+    void SetCursor(uint8_t, uint8_t);
+    void Clear();
```
stephenhensley, line 26: Add argument  names for and some (simple) doxygen reference documentation. 

for example:

```C++
/** Initializes the HDD44780 LCD display with the specified configuration */
void Init(const Config& cfg);
```

While the init function is fairly obvious, the others may not be as clear to the user. For example, I _assume_  `SetCursors` first argument is the x position, and the second argument is the y position, but I can't be sure without going all the way to the cpp file.

src/dev/lcd_hd44780.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+#ifndef DSY_LCD_HD44780_H
+#define DSY_LCD_HD44780_H
+
+#include "daisy_core.h"
+#include "per/gpio.h"
+
+
```
stephenhensley, line 8: Add a brief description in doxygen format to describe the class, and provide any specific hardware details that may not be immediately obvious to a user.

+1 for a url to an LCD display someone can order that'll work with this code.

stephenhensley (changes requested):
Overall looks good, only a few things to change. 😄 

Also, it looks like you can only enable/disable the cursor/blinking at Init time, but it might be nice to have functions for enabling/disabling that as well.

src/dev/lcd_hd44780.h
StaffanMelin: The SetCursor() is defined as SetCursor(iint8_t row, uint8_t col) so the first arg is row, the second is col.

I am sorry but I am very new to GitHub so I am not really sure what you want me to change or do! :)

src/dev/lcd_hd44780.h
StaffanMelin: https://wiki.dfrobot.com/Arduino_LCD_KeyPad_Shield__SKU__DFR0009_

StaffanMelin: I am not familiar with the doxygen format, could you provide me with a link? Sorry.

StaffanMelin: Do you want the functions to enable/disable blinking and the cursor right now, or could it wait for 2.0? :)

stephenhensley: @StaffanMelin Talking in multiple places at once can be a bit confusing. My bad.  In our direct messages I sent you a link to another libdaisy file  you can use as a reference for the doxygen stuff.

And yeah, I'm totally fine with skipping the enable/disable functions for now :)

stephenhensley:
Looks good! Thanks for adding the documentation. I'll merge it in a little bit 😄

stephenhensley (approved):

PR #311, opened 2021-03-03
SDMMC Update with DMA access

stephenhensley:
##  Overview

This reworks the internal bits of the SDMMC usage within libdaisy to work using the SDMMC's dedicated IDMA instead of working on a polling-based system, and adds configuration for common settings (clock powersave, clock speed, and buswidth for saving pins).

This should accomplish the following things:

1. The issues related to using the SD Card while audio is running should be  resolved.
2. The performance  of the SD Card via FatFS has been maximized to provide something usable for streaming audio, etc.
3. Configuration is available to support different bus widths for limited pinout devices, and speed for specific use cases or layouts.

### Available Configuration

Currently  SDMMC1 is the only supported peripheral, and the pin mapping is fixed.

BusWidth can be set to 4 bits or 1 bit (if saving pins is more important than bandwidth)

Speed has a  qualitative scale that maps to the following frequencies:

| Name | Frequency |
| --- | --- |
| SLOW | 400kHz |
| MEDIUM_SLOW | 12.5MHz |
| STANDARD | 25MHz |
| FAST | 50MHz |
| VERY_FAST | 100MHz |

**Note**: VERY_FAST setting is technically above the spec for 3v3-based transmission using SD Card. I have tested it on the Daisy Field, and found it to be quite stable, but ymmv.

### User-facing code changes 

This does break the previous `SdmmcHandler::Init()` function as it was. It adds a required Config struct, but that is the only breaking changes.

To adapt any previous code:

```cpp
// before:
sdhandler.Init();

// now:
SdmmcHandler::Config cfg;
cfg.Defaults();
sdhandler.Init(cfg);
```

### Other notes

The System clock tree has been edited to facilitate the frequencies. The only other thing affected by this was the SDRAM, which should still be operating  at 100MHz (the max frequency of the FMC). Though, some more testing and validation for this should be done.

I am still working through the following tests:

* [x] Reading all file info from root directory
* [x] Writing to a test file before audio starts
* [x] Reading back and verifying the written  file while audio is playing
* [x] Writing another file while audio is running.
* [x] Streaming audio directly from SD card
  * Confirmed smooth playback at default audio settings down to 400kHz 4-bit, or 12.5MHz 1-bit (400kHz 1-bit is too slow for streaming audio).
* [x] Writing audio to file while playing back audio (i.e. record a wav file).
* [ ] Read and Write to file with audio limitations (i.e. Can you do a looper that operates mostly off of the SD Card with working buffers).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,8 @@ per/i2c \
 per/spi \
 per/tim \
 per/uart \
-util/color
+util/color \
+util/WaveTableLoader \
 
 ######################################
 # building variables
--- a/core/Makefile
+++ b/core/Makefile
@@ -138,6 +138,20 @@ ifdef DAISYSP_DIR
 C_INCLUDES += -I$(DAISYSP_DIR)/Source
 endif
 
+# Include FATFS files
+# This does not include the additional option files
+# for Japanese, Simplified/Traditional Chinese, or Korean
+# since they would add an additional 480kB or so of flash
+FATFS_DIR ?= $(LIBDAISY_DIR)/Middlewares/Third_Party/FatFs/src
+FATFS_SOURCES ?=
+FATFS_SOURCES += \
+	$(FATFS_DIR)/diskio.c \
+	$(FATFS_DIR)/ff.c \
+	$(FATFS_DIR)/ff_gen_drv.c \
+	$(FATFS_DIR)/option/ccsbcs.c
+
+C_SOURCES += $(FATFS_SOURCES)
+C_INCLUDES += -I$(FATFS_DIR)
 
 # compile gcc flags
 ASFLAGS = $(MCU) $(AS_DEFS) $(AS_INCLUDES) $(OPT) -Wall -fdata-sections -ffunction-sections
--- a/libdaisy.vcxproj
+++ b/libdaisy.vcxproj
@@ -282,6 +282,7 @@
     <ClCompile Include="src\util\oled_fonts.c" />
     <ClCompile Include="src\util\sd_diskio.c" />
     <ClCompile Include="src\util\unique_id.c" />
+    <ClCompile Include="src\util\WaveTableLoader.cpp" />
     <ClInclude Include="Drivers\CMSIS\Device\ST\STM32H7xx\Include\stm32h750xx.h" />
     <ClInclude Include="Drivers\CMSIS\Include\cmsis_armcc.h" />
     <ClInclude Include="Drivers\CMSIS\Include\cmsis_armclang.h" />
@@ -485,6 +486,8 @@
     <ClInclude Include="src\util\scopedirqblocker.h" />
     <ClInclude Include="src\util\sd_diskio.h" />
     <ClInclude Include="src\util\unique_id.h" />
+    <ClInclude Include="src\util\WaveTableLoader.h" />
+    <ClInclude Include="src\util\WavWriter.h" />
     <ClInclude Include="src\util\wav_format.h" />
     <None Include="stm32.props" />
     <None Include="libdaisy-Debug.vgdbsettings" />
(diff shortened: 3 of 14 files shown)


stephenhensley:
Did quite a bit more testing, and am going to add a few utility classes (that are currently in the FieldSampler branch of the DaisyExamples repo). But everything seems to be working well. The utility classes so far are:

WaveTableLoader - loads a .wav file as wavetables and stores to internal RAM
WavWriter - can record a wav file from the audio callback (i.e. streaming audio to the SD Card).

The test I had running most recently would load a bank of wavetables from a .wav file, save it to RAM, and then allow you to record a .wav file (up to 32-bit 96kHz) of fiddling with the knobs of the wavetable synth. 

Saving files tested upto about 32MB so far, and there's no audible glitch in audio during the longer operations like `f_sync` or `f_lseek` . So I'm feeling pretty good about evertying.

stephenhensley:
Currently the templated WavWriter includes "fatfs.h" in libdaisy by default. This does add a bit more build complexity to every project. So we may want to change the format of how it's working. We could use a user-supplied buffer instead of a template (like with the WaveTableLoader.

recursinging:
I had problems with [`GPIO_InitStruct.Speed     = GPIO_SPEED_FREQ_VERY_HIGH;`](https://github.com/electro-smith/libDaisy/blob/444d175ddda0a1f91b271cfe1b7eac4ff50d69ed/src/per/sdmmc.cpp#L63) in sdmmc.cpp due to ringing when using a breadboard, (and my sub-optimal PCB routing).   This configuration affects the drive strength and slew rate of the GPIO, and I was only able to get an SD card functional on a breadboard by changing it to: `GPIO_SPEED_FREQ_LOW`.  The following configurations are available:

* `GPIO_SPEED_FREQ_LOW `
* `GPIO_SPEED_FREQ_MEDIUM`
* `GPIO_SPEED_FREQ_HIGH`
* `GPIO_SPEED_FREQ_VERY_HIGH`

It might be prudent to attach this GPIO speed configuration to the `SdmmcHandler::Speed` enum.  I haven't looked at the datasheet, so I don't know what effect these configs have on the GPIO rise time, but there is likely a sane correlation to the frequencies you've selected for the enum values.

recursinging:
I had a look at the Datasheet - Pages 133-134 outline the Frequencies and rise times to be expected at different capacitance and voltage.

If I understand the Datasheet correctly, the following mapping should cover most cases:

|  Speed  |  GPIO_SPEED  |
|-------------|----------------------|
| LOW       | GPIO_SPEED_FREQ_LOW|
| MEDIUM_SLOW | GPIO_SPEED_FREQ_MEDIUM |
| STANDARD | GPIO_SPEED_FREQ_HIGH |
| FAST | GPIO_SPEED_FREQ_VERY_HIGH|
| VERY_FAST | GPIO_SPEED_FREQ_VERY_HIGH|

I tested this pretty extensively a couple of months ago with a microsd card reader breakout on a breadboard. I was able to access the card reliably at 12.5MHz using the

stephenhensley:
@recursinging oh nice, I had seen that page a long time ago, and could not for the life of me remember where it was. Thanks for sharing! 

you're correct it looks like the mapping you mentioned should work. I'll be doing a patch soon to make the `Default` configuration part of the constructor for the Config struct. So I'll add this when I do that.

chrisfromwork:
Is there an expectation that WavWriter should work on patch? It looks like the default WavWriter setup doesn't call out needing to mount the sd card. When i mount the sd card currently, I typically see things hang when attemping a f_open for the file to save to

chrisfromwork:
seems like i had two problems:

1. WavWriter file operations need to be done in the main loop (Sample can be called on audio thread, but write and file open/save should be in main loop)
2. ExFat didn't work correctly for the sd card i had. Reformatting to FAT got things working.

FWIW, this sd card wave recorder is awesome. Thanks for adding this feature!

PR #314, opened 2021-03-18
Makefile: Added CPP_WARNINGS variable; Allowed user-configurable WARN…

TheSlowGrowth:
STM and CMSIS headers create lots of 
>warning: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
when compiling for C++17.

This commit adds `-Wno-register` to the C++ compiler calls.
To allow this, a new CPP_WARNINGS variable has been added.
Both CPP_WARNINGS and WARNINGS can now be pre-populated with user defined settings (`+=` instead of `=`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -281,7 +281,8 @@ C_INCLUDES = \
 -I$(MODULE_DIR) \
 -I. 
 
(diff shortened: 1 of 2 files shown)


stephenhensley:
checks out fine here.

PR #315, opened 2021-03-22
Fix compiler warnings on function return type const qualifiers.

kisielk:
This fixes a warning I was getting tired of seeing while compiling :)


stephenhensley:
makes sense. Thanks for the contribution!

PR #316, opened 2021-03-23
Allow user flags/defs in Makefile

grrrr:
Please allow user flags and user definitions in Makefiles.
Currently there is no (obvious) way to configure projects during make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,7 +256,7 @@ MCU = -mthumb $(FLOAT-ABI) $(FPU) $(CPU)
 AS_DEFS = 
 
 # C defines
-C_DEFS =  \
+C_DEFS +=  \
 -DCORE_CM7  \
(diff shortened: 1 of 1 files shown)


stephenhensley:
Ah, good point.

I believe many of these flags are already set up this way in the `core/Makefile` which is the project-level file. But I agree that this should be configurable for the library itself as well. 

Since this file doesn't include any others makefiles, what would be your desired way of configuring it other than editing the file itself?

grrrr:
Oh, i am sorry, my bad. I think i have had this lying around for some time, having overlooked the changes in core/Makefile. libdaisy does not have to be configured really – it would need to be modified anyway.

stephenhensley:
no problem! Usually the only thing I find myself tinkering with in the primary Makefile is the `DEBUG` setting.

PR #319, opened 2021-04-01
Fix MIDI parser for 2 byte message 

GregBurns:
Many MIDI keyboards send ChannelPressure messages when a key is held down for a while. This is a two byte message that gets the MIDI parser state machine out of synch such that the eventual Key-Up message is discarded. This patch handles ChannelPressure, the only 2 byte message in MIDI 1.0, as a special case and correctly queues it. A further enhancement would be to always resynchronize byt setting the state to ParserEmpty when the status bit is detected but ChannelPressure messages would still need to be special-cased.

This patch also adds a helper function for interpreting the PitchBend message.


stephenhensley:
Awesome! Thanks for the work. I'll have time to test this week.

stephenhensley:
Sorry for the delay!  Tests passed, and it's pitch bend is extra fun. Thanks again!

PR #323, opened 2021-04-13
Updates to UART

beserge:
Adds support for multiple UART periphs.
Moves UartHandler to pimpl pattern.
Adds config options.


src/per/uart.cpp
```diff
@@ -20,11 +24,40 @@ static UartRingBuffer DMA_BUFFER_MEM_SECTION   uart_dma_fifo;
...
+
+    int CheckError();
+
+    UartHandler::Result InitPins();
+
+    UartHandler::Result DeInitPins();
+
+    GPIO_TypeDef*  port;
+    USART_TypeDef* periph;
+    uint16_t       tx, rx;
```
stephenhensley, line 59: These shouldn't be necessary.

periph is accessible with huart.Instance without an additional pointer

port implies that only a single port can be used while both pins are below. This could lead to issues if tx/rx are on different ports.

The pins are already located in config_.pins, and I don't think they need to be stored for any particular reason.

src/per/uart.cpp
```diff
@@ -34,135 +67,371 @@ struct uart_handle
...
+                           uart7_pins_rx,
+                           uart8_pins_tx,
+                           uart8_pins_rx,
+                           lpuart1_pins_tx,
+                           lpuart1_pins_rx};
+
+bool gpiopin_equal(dsy_gpio_pin a, dsy_gpio_pin b)
+{
+    return a.port == b.port && a.pin == b.pin;
+}
```
stephenhensley, line 334: The `dsy_gpio_cmp` function in `daisy_core.h` already exists. You should be able to use that instead.

src/per/uart.cpp
```diff
@@ -215,106 +481,409 @@ void HAL_UART_AbortReceiveCpltCallback(UART_HandleTypeDef* huart)
...
+
+    for(int i = 0; i < 11; i++)
+    {
+        if(port == ports[i])
+        {
+            func_p[i]();
+            return;
+        }
+    }
+}
```
stephenhensley, line 566: This doesn't belong in uart.cpp, it should probably move to the hal_map file so that other peripherals can use it.

src/per/uart.cpp
```diff
@@ -215,106 +481,409 @@ void HAL_UART_AbortReceiveCpltCallback(UART_HandleTypeDef* huart)
...
 
 // HAL Interface functions
 void HAL_UART_MspInit(UART_HandleTypeDef* uartHandle)
 {
-    GPIO_InitTypeDef GPIO_InitStruct = {0};
-    if(uartHandle->Instance == USART1)
+    UartHandler::Impl* handle = MapInstanceToHandle(uartHandle->Instance);
+    GpioClockEnable(handle->port);
+    UartClockEnable(handle->periph);
```
stephenhensley, line 661: There's no 'reuse' of the UartClockEnable. So I almost think a simple if/else or switch over the peripheral instance within this function would be a bit simpler.  Same goes for the NvicEnable. 

The wrappers for each clock enable/disable is pretty long. And the only time those will need to get called is in the MspInit/DeInit functions.

The other aspect as that these aren't class functions so if we decide to add any flexibility to the interrupt priority, or even whether an interrupt gets enabled through the Config struct there would be no way of passing that along without more drastic changes.

src/per/uart.cpp
```diff
@@ -34,135 +67,371 @@ struct uart_handle
...
+        impl->rx_last_pos = cur_pos;
 #ifdef UART_RX_DOUBLE_BUFFER
         // Copy to queue fifo we don't want to use primary fifo to avoid
         // changes to the buffer while its being processed
         uint8_t processbuf[256];
-        uhandle.dma_fifo_rx->ImmediateRead(processbuf, len);
-        uhandle.queue_rx.Overwrite(processbuf, len);
+        impl->dma_fifo_rx->ImmediateRead(processbuf, len);
+        impl->queue_rx.Overwrite(processbuf, len);
 #endif
```
stephenhensley, line 435: I don't see any reason for this not to be a class function for the `Impl` class. You're basically making it that in a c-style way anyway.

You would just have the pointer to the Impl class call it's own RxComplete function instead of having the function have a pointer to the object (see sai.cpp for example).

stephenhensley (changes requested):
Haven't done any testing yet, but there's still a little bit of cleanup/reorg to do. Overall, looking good though!

stephenhensley (approved):

stephenhensley:
Everything here looks good. 

I'll merge tomorrow morning. So I can be around if any issues come up.

One thing I noticed while I was double checking if any examples have to change is that the `MidiHandler` class handles the UART initialization which is fine, because it should be able to configure the settings. It does feel a little weird not being able to pass any pins in.

We should consider updating the MIDI config to be similar to other so that alternate pins could be used. That said, MIDI still will only work for USART1. So that's kind of a mute point until we add DMA to the other peripherals. (Mostly just putting this here so it's written down _somewhere_).

PR #325, opened 2021-04-27
Remove redundant codec reset from patch

### LINKED ISSUE #320: DaisyPatch: External AK4556 Reset pin missing init
It looks like at some point the reset pin initialization for the second audio codec on the Daisy patch got removed from the firmware.

Just need to initialize the `ak4556_reset_pin_` with:

* mode = `DSY_GPIO_MODE_OUTPUT_PP`
* pull =`DSY_GPIO_NOPULL`
* pin = `DaisySeed::GetPin(29)`

beserge:
Remove the resets and the pin struct itself
Also opted to use the PIN_AK4556_RESET macro
rather than the hardcoded 29
Fixes #320

### DIFF
--- a/src/daisy_patch.cpp
+++ b/src/daisy_patch.cpp
@@ -40,10 +40,6 @@ void DaisyPatch::Init(bool boost)
     InitDisplay();
     InitMidi();
     InitControls();
-    // Reset AK4556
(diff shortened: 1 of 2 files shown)


stephenhensley:
Looks good!

PR #327, opened 2021-04-28
Feature/led driver bugfix

TheSlowGrowth:
Missing volatile qualifier resulted in this loop away being optimized into an infinite loop:
```
// wait for current transmission to complete
while(current_driver_idx_ >= 0) {};
```


stephenhensley:
Nice catch.

PR #330, opened 2021-04-29
Added const to input arrays in audio callbacks

### LINKED ISSUE #307: Mark the input of the audio callback as const
This came up while running lint checking (clang-tidy) on my code.
Any reason not to label the input of the audio callback with a const modifier?

From:
```cpp 
typedef void(* daisy::AudioHandle::AudioCallback) (float **in, float **out, size_t size)
typedef void(* daisy::AudioHandle::InterleavingAudioCallback) (float *in, float *out, size_t size)
```

To:
```cpp
typedef void(* daisy::AudioHandle::AudioCallback) (const float **in, float **out, size_t size)
typedef void(* daisy::AudioHandle::InterleavingAudioCallback) (const float *in, float *out, size_t size)
```

clang-tidy reference: [readability-non-const-parameter](https://clang.llvm.org/extra/clang-tidy/checks/readability-non-const-parameter.html)

TheSlowGrowth:
Fixes #307

### DIFF
--- a/src/hid/audio.h
+++ b/src/hid/audio.h
@@ -24,14 +24,16 @@ class AudioHandle
     };
 
     /** Non-Interleaving Callback format. Both arrays arranged by float[chn][sample] */
-    typedef void (*AudioCallback)(float** in, float** out, size_t size);
+    typedef void (*AudioCallback)(const float* const* in,
+                                  float**             out,
+                                  size_t              size);
 
     /** Non-Interleaving Callback format. 
      ** audio is prepared as { L0, R0, L1, R1, . . . LN, RN }
      */
-    typedef void (*InterleavingAudioCallback)(float* in,
-                                              float* out,
-                                              size_t size);
+    typedef void (*InterleavingAudioCallback)(const float* in,
+                                              float*       out,
+                                              size_t       size);
 
     AudioHandle() : pimpl_(nullptr) {}
     ~AudioHandle() {}


stephenhensley:
woah, it totally makes sense, but I have never seen this before:
```c++
const float* const* in
```
as the const representation of `float**`

Thanks for PR! This does require that _every_ DaisyExample projects gets updated. So I'll merge this once I have an extra few minutes to write up a little `sed` script to update them.

The only thing that I don't love about this is that it's another thing to potentially confuse newcomers to C++ or code in general.

GregBurns:
I think there is going to be a lot of casting back to float**. Is this really a good idea, the buffer can actually be modified with no ill-effects during the callback.

TheSlowGrowth:
> I have never seen this before

See here: https://stackoverflow.com/a/1143272

> potentially confuse newcomers

Newcomers would copy the function header for an audio callback from an example. Why would it confuse them more than the already confusing `float **`? If you would like to hide the type somewhat, we could also introduce a typedef
 like this `using AudioBlockPtr = const float * const *;` Then the audio callback function header would look a little cleaner to a newcomer.
 
> there is going to be a lot of casting back to float**

I don't see where that would be required. Can you give an example? Why would I want to modify the inputs directly?

> can actually be modified with no ill-effects during the callback

Sure, but it doesn't result in clean code if you modify your input array in the callback. In some of the projects I worked on professionally, we had lots of old code that would modify the function arguments during the execution. These kinds of functions are typically super nasty to debug and understand, especially once loops come into play. It's much easier to follow what the function is supposed to do if you can be sure that the function arguments have the same value everywhere in the function.

TheSlowGrowth:
Closing this due to #344

PR #332, opened 2021-04-30
Add a  CMake build system

claudiocabral:
Adds a CMake build system to libDaisiy.

Probably depends on this  PR: https://github.com/electro-smith/DaisySP/pull/153

Alongside CMakeLIsts.txt, a toolchain file was added to accomodate for the cross compiler.

In order to simplify building projects on top of cmake, mimicking the current make file structure, a `default_build.cmake` file was also added. Here's a gist examplifying how they work in conjunction:
https://gist.github.com/claudiocabral/04f436937423652dd6bbfe8ec4ada34c

I've tested this manually and it seems to "work on my machine" at least.
This PR attempts to follow current best practices in cmake builds for embedded systems


stephenhensley:
Fantastic, I think there have been quite a few discussions about this over the months. Excited to see some support begin to get added.

>I've tested this manually and it seems to "work on my machine" at least.

Hahahahah famous last words. 😄 

I've been a bit busy this week, but I'll try to get to checking this out and testing it on my machine over the next week or so. 

Thanks for the contribution!

claudiocabral:
glad to contribute, and happy to know it might help others :)
I got a patch a week ago, so looking forward to contribute and build on the platform

brbrr:
It works well in my tests. Although, I end up adding conditional to support both ARM and x64 builds:

```
if(${CMAKE_SYSTEM_PROCESSOR} MATCHES ARM)
	# include default_build.cmake etc
elseif(NOT ${CMAKE_SYSTEM_PROCESSOR} MATCHES ARM)
	# set subdirectories for DaisySP & portaudio, tests and add GoogleTest etc
endif()

```

Whole config file is here: https://gist.github.com/brbrr/d0a7c63db39569e9c7f31cd18e07888c

claudiocabral:
@brbrr I think what people tend to do is explicitly state their platform by setting the CMAKE_TOOLCHAIN_FILE variable
`cmake -B build -DCMAKE_TOOLCHAIN_FILE=cmake/toolchains/stm32h750xx.cmake`
This file takes care of setting up variables like `CMAKE_SYSTEM_PROCESSOR`.

The main advantage is that you can add new arbitrary new targets without touching the existing build system.

Of course, in the case of a particular build like yours (aka, you know your target is ARM but you want to run tests with on x86_64), it makes perfect sense to specialize a bit

brbrr:
>  explicitly state their platform by setting

Yeah, that's what I did in one of the conditional branches.

I'd love to know more about other's workflows, but from where I stand, it feels natural/easier to develop DSP code on my machine, completely separate from hardware. And then integrate it with daisy controls etc via another build.

cmake/default_build.cmake
```diff
@@ -0,0 +1,48 @@
+add_subdirectory(${LIBDAISY_DIR} libdaisy)
+add_subdirectory(${DAISY_SP_DIR} DaisySP)
```
TheSlowGrowth, line 2: Can we make DaisySP optional?

TheSlowGrowth:
I just wanted to chime in and say that I LOVE this PR. Please, please merge it!
I managed to setup an [automated hardware testing project](https://github.com/TheSlowGrowth/daisyHat) with CMake in just a couple of hours and I'm NOT a CMake expert by any means. The files provided here worked fine for my and I'd love to see this become a supported way of using libDaisy.

(Also: Imagine if we used CMake as the default build system - it would make things easier for so many beginners. No need to manually build libDaisy BEFORE building the user project - CMake does that automatically for you)

cmake/default_build.cmake
claudiocabral: this default build is supposed to work as a copy of the `Makefile` build;
you can also use it as a reference to make your own :)

claudiocabral:
@TheSlowGrowth thanks for the support :)
updated the PR to work after the last couple of merges

cmake/default_build.cmake
TheSlowGrowth: The Makefile accepts if you don't have DAISY_SP_DIR set - it will not link against daisySP then

claudiocabral: hm, didn't notice that; will add the check then :)

stephenhensley:
Sorry for the huge delay on this one. Overall, it looks solid; I don't see any glaring issues. Good catch on ensuring DaisySP is optional. I'll try to get it merged in the next few days.

I agree with @TheSlowGrowth that we could very well see CMake becoming the default build system in the future. It is infinitely easier to manage the cross platform builds for projects that use DaisySP. I've exclusively been using CMake for most non-hardware projects lately, and prefer it to Make in nearly every way.

claudiocabral:
updated to check if `DAISYSP_DIR` is set before including DaisySP.
Also moved some compiler flags to the toolchain file.
hope we can get it merged soon :) As always, let me know if there's anything else that's required

stephenhensley:
I've been wanting to take the time to make a project or two with this to try it out first, but I haven't been able to find the time.

I don't see any issues, and since it has worked for @claudiocabral and @TheSlowGrowth I'm sure it's in good shape.

We're going to be making a separate wiki page for creating new projects pretty soon (most of which will be based around the current Make build, and using the helper.py script in DaisyExamples), but it would be really great to have a tutorial/guide for people setting up CMake projects as well. Feel free to reach out if you have any interest on helping out making that.

claudiocabral:
happy to see this merged in! Happy to contribute with the wiki when the time comes

PR #334, opened 2021-05-04
Ensure clang-format is called on files in src/

recursinging:
`fix_style.sh` invokes `clang-format` on files in subdirectories of `src/`, but not files located directly in `src/`.  This is just a small fix to include them.


stephenhensley:
Great. I had missed that it wasn't catching this. Guess there haven't been a lot of changes to the board support files since the folder reorganization. 

Thanks!

PR #337, opened 2021-05-05
Blocksize switch bugfix

### LINKED ISSUE #297: ADC averaging not tracking audio callback rate changes
The ADC controller initialization uses the default audio frequency and blocksize to compute the slew parameters. If these are subsequently changed the slew rate is no longer valid and can be off by a lot. For example, the default audio callback frequency is 1KHz but with an audio block size of 256 at 32KHz sample rate, the callback frequency is only 125Hz. The controller Init() api is private so calling that after changing the sample rate and block size is not an option.

### LINKED ISSUE #317: Switch::TimeHeldMs() does not reflect real milliseconds when blocksize is not 48
Also has an effect on the Encoder function of the same name through composition.

The sample rate is known. So it  should be an easy fix to convert from samples to milliseconds. This was just an oversight from when block size was hard coded.

beserge:
Resets update rates for HIDs when board change samplerate or block size. 
Tested on all boards except version.

Small note: some HIDs refer to the update rate as 'SampleRate', and others refer to it as 'UpdateRate'.
Should probably unify that across classes. My vote goes for UpdateRate.

Fixes #317 
fixes #297

### DIFF
--- a/src/daisy_field.cpp
+++ b/src/daisy_field.cpp
@@ -186,9 +186,19 @@ void DaisyField::ChangeAudioCallback(AudioHandle::AudioCallback cb)
     seed.ChangeAudioCallback(cb);
 }
 
+void DaisyField::SetHidUpdateRates(){
+    //set the hids to the new update rate
+    for(size_t i = 0; i < SW_LAST; i++)
+    {
+        sw[i].SetUpdateRate(AudioCallbackRate());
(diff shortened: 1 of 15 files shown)


stephenhensley:
Looks good! Verified on Daisy Pod, and made sure other board files are handled in the same fasion.

PR #338, opened 2021-05-06
Correct AudioHandle blocksize check, fixes #306

### LINKED ISSUE #306: AudioHandle::SetBlockSize check is incorrect
The buffer size passed to the SAI StartDMA function is
 `blocksize * 2 * 2` 
so I believe the maximum block size check should be
 `kAudioMaxBufferSize / 4`   

Proposed fix:

`AudioHandle::Result SetBlockSize(size_t size)`
`{`
`     size_t maxSize = kAudioMaxBufferSize / 4;`
`     config_.blocksize = size <= maxSize ? size :  maxSize;`
`     return size <= maxSize ? AudioHandle::Result::OK :` 
`                             AudioHandle::Result::ERR;`
`}`

beserge:
Tested on many blocksizes, all working.
Confirmed large blocksizes  (~1000) would cause issues on master.
fixes #306

### DIFF
--- a/src/hid/audio.cpp
+++ b/src/hid/audio.cpp
@@ -51,10 +51,10 @@ class AudioHandle::Impl
 
     AudioHandle::Result SetBlockSize(size_t size)
     {
-        config_.blocksize
-            = size <= kAudioMaxBufferSize ? size : kAudioMaxBufferSize;
-        return size <= kAudioMaxBufferSize ? AudioHandle::Result::OK
-                                           : AudioHandle::Result::ERR;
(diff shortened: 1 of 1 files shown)


stephenhensley:
Confirmed this check makes sense with how it's currently set up. I am having a hard time remembering why the second `*2` is necessary (the first is to account for the double buffering).

This does reinforce the fact the the current actual maximum block size (that you can pass into the audio driver) is only 256 samples. So we should consider making those static buffers a bit larger (at least so 1024 samples are possible per callback). However, that's definitely outside the scope of this PR. So we can do that in another PR.

PR #339, opened 2021-05-08
Add abstract base class to OledDisplay class

TheSlowGrowth:
## What does this PR change?
This PR adds an abstract base class `OneBitGraphicsDisplay` to `OledDisplay`, shifting all of the existing drawing functions to this base class.

## Why is this PR a good idea?

1. Binary size

`OledDisplay<..>` is a template class. If multiple different types of displays are used in a project, e.g. `OledDisplay<SSD130xDriver<128, 64, SSD130x4WireSpiTransport>>` and `OledDisplay<SSD130xDriver<128, 32, SSD130xI2c128x32Driver>>`, all of the drawing functions end up multiple times in the final binary, even though they do exactly the same thing. This is because each instantiation of the template `OledDisplay<..>` is techincally a completely independent class.

Having the drawing functions in a common abstract base class resolves this problem, because all versions of `OledDisplay<..>` use the same functions from their common base class `OneBitGraphicsDisplay`

2. Code modularity

Without a common base class, each version of the template `OledDisplay<..>` is techincally a completely independent class. This makes it impossible to write common drawing functions, like this: 
```cpp
void drawSlider(OledDisplay<...>& display, float value, ....);
void drawCheckBox(OledDisplay<...>& display, bool isTicked, ....);
```
Of course, those drawing functions could be templated as well, but then the issue of binary size bloat becomes even worse.
With a common base class, the drawing functions could look like this:
```cpp
void drawSlider(OneBitGraphicsDisplay& display, float value, ....);
void drawCheckBox(OneBitGraphicsDisplay& display, bool isTicked, ....);
```
... making it possible to not only reuse them for all versions of `OledDisplay<..>` but also for any other 1bit-per-pixel display class that someone may want to write.

3. No API change

User projects are not affected. The current API stays intact.

## Possible drawbacks

Now that many of the drawing functions are `virtual`, we may see a slight loss of drawing performance in drawing functions that call `DrawPixel()` very often.

If this should turn out to be problematic, a possible solution would be to overwrite these drawing functions again in `OledDisplay<..>` and make these draw to the pixel buffer directly, avoiding the virtual calls to `DrawPixel()`. Another solution could be to provide auxiliary drawing functions like `DrawFastHorLine()` and `DrawFastVertLine()` to reduce the number of function calls required to draw certain shapes.

## Outlook

I request this change to be able to provide a generic menu system that allows users to quickly setup complex menu structures without having to implement the actual drawing routines for their particular display type. Having generic drawing routines that directly draw to a `OneBitGraphicsDisplay` makes it possible to implement this menu system without having to provide an individual version for EACH AND EVERY combination of display driver and transport.


TheSlowGrowth:
@recursinging just notifying you here as you recently made changes to this code as well.

recursinging:
Thanks for the notification, this was under my radar. 

I agree with your arguments for the most part, it's likely rare to drive multiple displays, so the binary size argument alone is pretty weak. I find the mentioned code modularity is a very relevant point. As a matter of fact, I was recently discussing with @grrrwaaa how to best handle differences in display dimensions (in that case, the smaller kxmx_bluemchen display) when implementing UI elements with oopsy.  This PR would enable higher levels of abstraction and thus allow for more robust, flexible, and re-usable components, rather than every application having to push pixels and re-inventing the wheel.  While perhaps not as much of a factor for these little OLED displays, having higher levels of abstraction becomes increasingly important as screen real-estate increases.  Having the option to build useful libraries on top of lower abstractions makes a lot of sense.

Without benchmarking the performance cost of virtual method invocation, but simply judging by its use in the [Adafruit GFX library](https://github.com/adafruit/Adafruit-GFX-Library), and their support for low spec MCUs, I'd wager it won't outweigh the benefit of the abstraction.

I agree with these changes, and would like to see this PR merged.

stephenhensley:
I agree with @recursinging that the binary-size argument isn't very compelling. It would be rare to drive multiple displays from a single daisy, and it doesn't look like the code size would be too much larger to run 2 different displays (However, 

Portability is a much stronger case for this, and has me mostly persuaded. 

Have you looked into what this might look like using [CRTP (Curiously Recurring Template Pattern)](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) as a compile time replacement for this type of reusable implementation. 

I agree that no matter what, this type of interface needs to be abstract and flexible enough to work with various displays without tons of redundant copy/paste. A (small) performance hit would be worth the maintainability in that case. I'd just like to ensure that we've explored compile-time alternatives that wouldn't result in the (possibly slight) performance hit. Also, _most_ display usage I've seen so far (aside from o-scopes, etc.) could be updated at really low framerates, and there are very few pixels compared to larger displays. So performance may not be a huge deal breaker for UIs built this way.

So all that said, I think I'm onboard, though it should be made clear that this pattern should not be used on a regular basis, and does incur some runtime overhead. If we had all the free time in the world I'd love to do some benchmarking and see what the virtual call overhead actual costs in terms of CPU time on a real project.

Also, semi-related since this does touch the various Width/Height getters, would the virtualization of those methods prevent compile-time access to the Height/Width (as requested in #342) or does the `const` attribute still allow the compiler to essentially de-virtualize that value during compile time?

If CRTP is a no-go, then I think this looks like it's in a mergeable state.

TheSlowGrowth:
What would a CRTP look like when applied to the Display class?
```cpp
class OledDisplayA : public OneBitGraphicsDisplay<OledDisplayA>
{ 
/* ... */ 
};
class OledDisplayB : public OneBitGraphicsDisplay<OledDisplayB>
{ 
/* ... */ 
};

OledDisplayA dispA;
OledDisplayB dispB;
OneBitGraphicsDisplay* displayPointer1 = &dispA; // doesn't work!
OneBitGraphicsDisplay* displayPointer2 = &dispB; // doesn't work either!
```

What we want here is a way to be able to assign `dispA` and `dispB` to the same pointer type so that we can pass both of them into the same drawing function. CRTP makes both `OledDisplayA` and `OledDisplayB` inherit from the same base template class, but different template instantiations of that base class. Hence both of them are totally independent classes and can't be assigned to the same pointer type. I've used CRTP a lot in other situations, but it's only useful when you need to call stuff in the derived class from inside the base class. And that's not the problem we're trying to solve here.

Our situation (an interface and multiple implementations of it) is exactly what virtual functions are designed for. Why work around that? Or asking differently: What other options do we have anyway? 

However, I have an idea how we can remove many of the virtual function calls: We can use the qualified name of functions to explicitly suppress the virtual dispatch mechanism, like this:
```cpp
OledDisplay* disp = getDisplayPointerFromSomewhere();
disp->OledDisplay::DrawPixel(x, y, on); // direct function call
disp->DrawPixel(x, y, on); // virtual function call
```
Then how about this solution with an intermediary, CRTP class:
```cpp
class OneBitGraphicsDisplay
{
public:
    virtual void DrawPixel(int x, int y, bool on) = 0;
    virtual void DrawLine(int x1, int y1, int x2, int y2, bool on) = 0;
    virtual void DrawRect(int x1, int y1, int x2, int y2, bool on, bool fill) = 0;
    // ...
};

template<class ChildType>
class OneBitGraphicsDisplayImpl : public OneBitGraphicsDisplay
{
    // We implement all the higher level drawing functions here.
    // Wherever we need to call a lower level function, e.g. 
    // DrawPixel(), we can use the qualified function name
    // to avoid a virtual function call.

    void DrawLine(int x1, int y1, int x2, int y2, bool on) override
    {
        // ...
        ChildType::DrawPixel(x, y); // direct call
    }
    void DrawRect(int x1, int y1, int x2, int y2, bool on, bool fill) override
    {
        // ...
        ChildType::DrawLine(x1, y1, x2, y1); // direct call
        ChildType::DrawLine(x2, y1, x2, y2); // direct call
        ChildType::DrawLine(x2, y2, x1, y2); // direct call
        ChildType::DrawLine(x1, y2, x1, y1); // direct call
    }
};

template<typename DisplayDriver>
class OledDisplay final : public OneBitGraphicsDisplayImpl<OledDisplay<DisplayDriver>>
{
public:
    // this function will only be called virtually, if we do
    //      OneBitGraphicsDisplay* disp;
    //      disp->DrawPixel(...);
    // But if we call ...
    //      OledDisplay<Driver>* disp;
    //      disp->DrawRect(...);
    // ... then only the DrawRect() call itself will be virtual,
    //  the internal calls to DrawPixel(), DrawLine(), etc. will be non-virtual.
    void DrawPixel(int x, int y, bool on) override 
    {
        // ...
    }
};
```
Note that `OledDisplay` is marked `final` to prevent further inheritance. This is to avoid problems in child classes. Let's assume that someone wants to just replace the `DrawLine()` function like this:
```cpp
class MyCustomizedOledDisplay : public OledDisplay<...>
{
public:
    void DrawLine(...) { /* some custom line drawing */ };
};

MyCustomizedOledDisplay disp;
disp.DrawLine(...); // calls MyCustomizedOledDisplay::DrawLine as expected
disp.DrawRect(...); // calls OledDisplay::DrawRect() which only uses OledDisplay::DrawLine() internally
```
But I think that's a good tradeoff.

stephenhensley:
Ah, good point! Totally spaced on the fact that you'd have to pass the specific templated OledDisplay type into the drawing functions, which defeats the whole point.

Thanks for the thorough response and immediate suggestions!

Let me look this over a bit and get back to you, but being able to suppress the virtual function call with the direct call seems like a good way to avoid the performance hit for those that want to. 

And worth mentioning that I don't want to overcomplicate it by just throwing another design pattern in there, but if it performs better/faster, and doesn't increase complexity on the user-end then I'm all for it.

TheSlowGrowth:
I implemented the CTRP-based intermediary class. Here's a quick test I made:
```cpp
#include <daisy.h>
#include <dev/leddriver.h>
#include <dev/oled_ssd130x.h>

using OledDisplayTransportType = daisy::SSD130x4WireSpiTransport;
using OledDisplayDriverType = daisy::SSD130xDriver<128, 64, OledDisplayTransportType>;
using OledDisplayType = daisy::OledDisplay<OledDisplayDriverType>;

OledDisplayType disp;

void testFunction()
{
    disp.DrawRect(0, 1, 2, 3, true, false);
}
```

I compiled that with `-O3` and looked closely at the disassembly. `DrawRect()` internally calls `DrawPixel()` and `DrawLine()`.  It seems to me as if it worked and these calls are now inlined :-) In the disassembly, I can cleary see lines like these:
```
buffer_[x + (y / 8) * width] |= (1 << (y % 8));
```
which come from the SSD1306 driver code. Apparently, inside of `DrawRect()` gcc inlines `OledDisplay<OledDisplayDriverType>::DrawPixel()` which in turn inlines `daisy::SSD130xDriver<...>::DrawPixel()`. So: Mission accomplished!

I was also curious about how much performance we gained with this.
Here's a little test code that I made:
```cpp
void displayTest(daisy::OneBitGraphicsDisplay* disp)
{
    volatile auto start = daisy::System::GetNow();
    for (int i = 0; i < 50000; i++)
    {
        disp->DrawArc(32, 32, 16, 0, 180, true);
        disp->DrawRect(32, 32, 16, 2, true, true);
        disp->WriteString("My Dummy Test", Font_7x10, true);
    }
    volatile auto end = daisy::System::GetNow();
    volatile auto diff = end - start;
}
```
I ran this code with the CRTP-based intermediary class (302ms) and with the "virtual only" solution that we had before (309ms). 
Yea, not much difference at all. I guess compared to all the calculations in DrawPixel() alone (lots of array adressing, bitwise logic, MODULO operators, loops, etc.) the virtual function calls don't make such a massive difference after all.

stephenhensley:
Nice! Thanks for giving it a try, and doing the performance test!

Great to know that it is indeed a pretty slim performance hit, and furthermore even better that we can avoid that hit anyway.

I think I'm signing off on it as-is unless there's anything else you'd like to change.

Thanks for all the hardwork! 🎆 

---

Final question (and I can check this a little bit later myself if you don't know of the top of your head, also this could easily be done in a future PR if it's not possible at the moment):

If I wanted to have a custom buffer outside the class for a display (i.e. a `uint8_t buffer[width * height]`) for a pregenerated animation or something. I'd probably want access to the width/height at compile time so I can statically allocate some space for it. With the current implementation it looks like this isn't possible. Since the height and width are template arguments, they're known at compile time. So is there a common way to expose these up to the `OledDisplay` class? My first thought, was just making them `constexpr` but figured I'd ask before I start trying anything.

Mostly asking because I think oopsy had code that relied on knowing the height/width at compile time (with the macros that used to be there). And there've been some macros hacked in to replace the old ones for now, but if there was an in-class method it would be convenient.

TheSlowGrowth:
constexpr functions cannot be virtual at the same time, at least for C++17 and below (C++20 allows it afaik, but there may be restrictions that I don't know of).

With that in mind, you won't be able to use the virtual functions `Width()` and `Height()` in a constexpr context. However, in your case, you don't have to use the `OneBitGraphicsDisplay` interface, as you probably know the actual type of the display `OledDisplay<...>`. So nothing stops us from adding a constexpr pair of getter functions, OR a two static members like this: 
```cpp
template<DriverType>
OledDisplay: public OneBitGraphicsDisplayImpl<OledDisplay<DisplayDriver>>
{
public:
    static constexpr width = DriverType::Width();
    static constexpr height = DriverType::Height();
};
using OledDisplayType = OledDisplay<...>;
uint8_t myBuffer[OledDisplayType::width * OledDisplayType::height];
```

stephenhensley:
@TheSlowGrowth noted! I'll copy that to the issue so that myself, or someone else can add it.

I'm signing off on this, especially with all of the thorough documentation for how/why it works, and avoiding unnecessary virtual calls. It looks great. Thanks again for the wonderful contribution!

PR #340, opened 2021-05-08
Feature/Unit Tests and constexpr FixedCapStr

TheSlowGrowth:
This PR adds a ton of stuff from my [Wunderkiste project](https://github.com/TheSlowGrowth/Wunderkiste)

- A unit test makefile and build target within libDaisy. This builds unit test on the local machine - it's not on the hardware but much better than no tests at all
- a github action for the unit tests
- vscode integration files for building, debugging and testing
- A fixed capacity string class for statically-allocated, safe, convenient and constexpr string manipulation.


tests/gtest_main.cpp
```diff
@@ -0,0 +1,44 @@
...
+        auto str = std::string(argv[i]);
+        if(str.rfind("--gtest_output", 0) == 0)
+        {
+            // this is the path we need to modify
+            // replace drive letter, then replace all backslashes
+            str = std::regex_replace(
+                str, std::regex("([A-Z]):\\\\"), "/cygdrive/$1/");
+            str = std::regex_replace(str, std::regex("\\\\"), "/");
+
+            printResult = str;
```
TheSlowGrowth, line 24: Note to self: Remove this print stuff that's left over from previous debugging

polyclash:
Google Test dependancy it shoul be optional.

TheSlowGrowth:
How would you add unit tests then? Any testing framework adds a dependency. A selfmade test runner is not maintainance-free either.
You can still build libDaisy without even checking out the googletest submodule. In fact, linDaisy itself doesn't include any googletest code at all. The googletest makefile builds a completely separate binary that just happens to include some of the files from libDaisy as well.

TheSlowGrowth:
I realised I should probably clarify a bit more what I'm trying to achieve with this PR.

### Why are unit tests a good thing for libDaisy
From my experience it is immensly helpful to have automated tests for a large codebase. I personally see four benefits:
1. When you have a high test coverage, making changes to existing code is a much more pleasant experience. If the tests are green after you made some changes, you can feel safe.
2. In writing the tests for new code, you force yourself to really think about what your code is supposed and not supposed to do. This helps you split a new functionality into small, independent parts that serve a well defined purpose. Additionally, the test itself is the best documentation. It clearly states how the code should behave.
3. Unit tests give you a great, simple environment to quickly run and test your code with a debugger. Not only does a test work independently from the hardware, it also gives you a clean, blank slate that's easy to setup so that you can quickly check out a particular scenario. Want to quickly try out that idea you had earlier? Add a *.cpp file to the `tests/` directory and add a simple test. Compile, Run, Debug. Compare that to a workflow without unit tests. You'd have to setup a hardware testbed whenever you want to actually try something.
4. When you write new code with unit tests in mind, the code tends to become more modular, reusable and clean. Automated tests require you to design your APIs in a way that lets you "fake" the behaviour of other pieces of software that your code interacts with. This gently pushes you towards cleaner interfaces.

In libDaisy, we don't yet have any way to write automated tests. That's what I'm trying to address with this PR.

### How the unit tests are setup
I added a new folder `tests/`. Inside of this folder, there's a makefile that compiles a commandline program for your computer. This is the test runner that runs the tests on your local machine. This is also what's compiled and run when you make a pull request. You can run this program manually if you like (it prints the test results to the console). The much nicer way is to use Visual Studio Codes Test panel (the image shows python code, but it works the same with C++): 
Test panel in VS Code
Here you can quickly run and debug individual tests, test suites or all tests combined.

Inside of the `tests/` folder, you can simply add a new `*.cpp` file to have it automatically built as part of the unit test binary. Thats literally all that's needed. I added the neccessarry Visual Studio Code integration files so that you can compile and run everything quickly without any further setup. 

### Testing example: The FixedCapStr class
The Fixed capacity string class (no dynamic memory allocation) is a convenient way to work with strings without having to manually declare buffers and call error-prone C functions on them. I had planned for a while to add this to libDaisy as it comes in very handy when working with files or displays. With the testing in place, I figured it would be a great opportunity to show how unit tests could be used in libDaisy.

Please check out [FixedCapStr_gtest.cpp](https://github.com/electro-smith/libDaisy/blob/3981b8bb15fc7a34aa1d56ddbb3ae1b18790f15c/tests/FixedCapStr_gtest.cpp) to see what a suite of tests could look like.

### What comes next
I would encourage anyone to try unit testing. I know that it won't be possible for everything (e.g. anything that uses actual peripherals of the hardware won't be testable). However, there are many classes and functions that don't actually need hardware to do their thing, e.g. the afforementioned string class. 

**Should this be merged, I would volunteer to write a quick tutorial on how to get started with unit testing in libDaisy**. It's really much simpler than it sounds.

TheSlowGrowth:
Actually, I just added a small tutorial to the documentation, see commit "Added a unit testing guide".

stephenhensley:
@TheSlowGrowth I agree 100% that tests are a good thing, and make things a lot easier, and safer to change.

I also agree that the tests should _not_ be required, but should be encouraged. That also opens up a new avenue for contribution for anyone who wants to help out by writing tests.

The addition seems solid, and I like the `FixedCapStr` concept. Definitely seems a bit easier to deal with than c strings, and doesn't carry the baggage of `std::string`. 

Thanks in advance for adding a guide for adding them. I don't see it anywhere, can you share a link to the commit you made. I'd love to check it out.

I've been a bit busy this week, which is why I've been slow to respond here. But this looks mergeable. I want to look over the `FixedCapStr` a bit before pulling the trigger (but the unit test definitely makes that a bit easier 😄 ). I should be able to look it over some time tomorrow and get this merged.

---

WRT to testing peripheral code, etc. It would be great to eventually add some (probably manual-only) tests for the various peripherals that could work with a pair of Daisy Seeds (since most peripherals can be tested by connecting to the opposite mode of the same peripheral on another DaisySeed). This is outside the scope of this PR, but definitely a possible solution to not having tests for peripherals.

TheSlowGrowth:
Hi Stephen!
The guide is in one of the commits of this PR!

Yes, a hardware test would be awesome. Imagine we had a little hardware setup with two daisy seeds connected via I2C, SPI and UART. Then a raspberry pi that listens to webhooks somehow (maybe a small Jenkins server?). Whenever it receives a request, it builds various test programs and runs them in the two seeds, then checks the result and reports it back here in the pull requests. Ahhhh that would be awesome but SO MUCH WORK to setup...

stephenhensley:
@TheSlowGrowth oh duh! oops! Sorry for the noise. I'll give it a look!

And indeed! A good long-term pipedream hahaha, but in the shorter term we could work toward just having tests that are able to be manually run with one or two seeds (for example UART could be tested between multiple peripherals on one Daisy, while SAI could be tested with a pair of daisies connected to one another). Just an idea. We can open up an issue to discuss it further.

stephenhensley:
Okay, fantastic! Proof of value for the unit-tests was pretty immediate reviewing this. Since I could really just rely on the the test to make sure the `FixedCapStr` class worked as expected.

The footprint for the tests is exceptionally small (one optional submodule, and the tests themselves). So that's great.

Thanks for adding the guide. I think it's thorough, and definitely gives a jumping off point for anyone interested in adding test (or simply running them) without being intimidating or difficult.

I didn't verify the Cygwin steps, but gave a quick demo run of building/running the tests from WSL (plus they were already run through CI).

Thanks for yet another wonderful contribution, @TheSlowGrowth!

PR #341, opened 2021-05-11
I2c slave functionality

CorvusPrudens:
Added slave functionality to libdaisy i2c. By extension, the master can also receive transmission with DMA on the off-chance that might be useful.

A couple private methods were renamed, but no public methods have changed, and no code should break as a result of these changes.

~Gabriel


src/per/i2c.cpp
```diff
@@ -47,11 +54,12 @@ class I2CHandle::Impl
     // scheduling and global functions
     struct DmaJob
     {
-        uint16_t                       slave_address    = 0;
+        uint16_t                       slave_address    = 0x10;
```
TheSlowGrowth, line 57: what's the reason for this change?

src/per/i2c.h
```diff
@@ -125,8 +143,34 @@ class I2CHandle
...
 
+    /** Receives data with a DMA and returns immediately. Use this for larger transmissions.
+     *  The pointer to data must be located in the D2 memory domain by adding the 
+     *  `DMA_BUFFER_MEM_SECTION` attribute like this:
+     *      uint8_t DMA_BUFFER_MEM_SECTION my_buffer[100];
+     *  If that is not possible for some reason, you MUST clear the cachelines spanning the size of 
+     *  the buffer, before initiating the dma transfer by calling 
+     *  `dsy_dma_clear_cache_for_buffer(buffer, size);`
+     * 
+     *  A single DMA is shared across I2C, I2C2 and I2C3. I2C4 has no DMA support (yet).
```
TheSlowGrowth, line 154: `I2C1` instead of `I2C` ?

src/per/i2c.cpp
```diff
@@ -303,20 +355,71 @@ I2CHandle::Result I2CHandle::Impl::ReceiveBlocking(uint16_t address,
...
+    HAL_StatusTypeDef status;
+    if(config_.mode == I2CHandle::Config::Mode::I2C_MASTER)
+    {
+        status = HAL_I2C_Master_Receive(
+            &i2c_hal_handle_, address << 1, data, size, timeout);
+    }
+    else
+    {
+        status = HAL_I2C_Slave_Receive(&i2c_hal_handle_, data, size, timeout);
+    }
```
TheSlowGrowth, line 367: I would leave out the `{}`

src/per/i2c.cpp
```diff
@@ -371,46 +478,128 @@ I2CHandle::Impl::StartDmaTransfer(uint16_t                       address,
...
+    if(HAL_DMA_Init(&i2c_dma_tc_handle_) != HAL_OK)
     {
         Error_Handler();
     }
 
-    __HAL_LINKDMA(&i2c_hal_handle_, hdmatx, i2c_dma_tx_handle_);
+    __HAL_LINKDMA(&i2c_hal_handle_, hdmatx, i2c_dma_tc_handle_);
 
     // start the transfer
     // TODO: Add ScopedIrqBlocker
```
TheSlowGrowth, line 515: Could you secure this block (and all other lines that access these variables that are shared between a IRQ and the "main thread") with a `ScopedIrqBlocker` and remove this comment? 
I think when this code was written, we didn't yet have the `ScopedIrqBlocker`, hence the comment.

TheSlowGrowth (changes requested):
These are some wonderful cahnges, good job!
I haven't yet tested the changes on actual hardware but the code look good to me. I'll try to test it out on hardware later

src/per/i2c.cpp
CorvusPrudens: 0 is a reserved address for I2C (as are 1-7 and 119-127), so 16 is a safer default address. Not a big change, and I could revert it, but it feels a bit more appropriate this way.

src/per/i2c.h
CorvusPrudens: Yeah I saw that and I'm not sure why I didn't correct it. Guess I forgot!

TheSlowGrowth (approved):

stephenhensley:
Thanks @CorvusPrudens for the great work! And thanks @TheSlowGrowth for jumping on the review so quickly!

Everything's passing, and the requested changes have been made. Overall, it's looking pretty solid to me. I'm going to carve out some time tomorrow to hook up a pair daisy seeds and do some physical testing, but if everything checks out this should be good to go!

stephenhensley:
Works like a charm! Tested at a few different clock speeds, etc.

Great job! Thanks for writing a test as well!

PR #344, opened 2021-05-17
Change callback input to const float

### LINKED ISSUE #307: Mark the input of the audio callback as const

beserge:
fixes #307

### DIFF
--- a/src/hid/audio.cpp
+++ b/src/hid/audio.cpp
@@ -387,7 +387,7 @@ void AudioHandle::Impl::InternalCallback(int32_t* in, int32_t* out, size_t size)
                 break;
             default: break;
         }
-        cb((InputBuffer)fin, (OutputBuffer)fout, size / 2);
+        cb(fin, fout, size / 2);
         // Reinterleave and scale
         switch(bd)
         {
--- a/src/hid/audio.cpp
+++ b/src/hid/audio.cpp
@@ -387,7 +387,7 @@ void AudioHandle::Impl::InternalCallback(int32_t* in, int32_t* out, size_t size)
                 break;
             default: break;
         }
-        cb((const float**)fin, (float**)fout, size / 2);
+        cb((InputBuffer)fin, (OutputBuffer)fout, size / 2);
         // Reinterleave and scale
         switch(bd)
         {
--- a/src/hid/audio.cpp
+++ b/src/hid/audio.cpp
@@ -387,7 +387,7 @@ void AudioHandle::Impl::InternalCallback(int32_t* in, int32_t* out, size_t size)
                 break;
             default: break;
         }
-        cb((float**)fin, (float**)fout, size / 2);
+        cb((const float**)fin, (float**)fout, size / 2);
         // Reinterleave and scale
         switch(bd)
         {
--- a/src/hid/audio.h
+++ b/src/hid/audio.h
@@ -23,15 +23,22 @@ class AudioHandle
         ERR,
     };
 
+    typedef const float* const* InputBuffer;
+    typedef float*              OutputBuffer;
+
     /** Non-Interleaving Callback format. Both arrays arranged by float[chn][sample] */
-    typedef void (*AudioCallback)(const float** in, float** out, size_t size);
+    typedef void (*AudioCallback)(InputBuffer  in,
+                                  OutputBuffer out,
+                                  size_t       size);
+
+    typedef const float* InterleavingInputBuffer;
 
     /** Non-Interleaving Callback format. 
      ** audio is prepared as { L0, R0, L1, R1, . . . LN, RN }
      */
-    typedef void (*InterleavingAudioCallback)(const float* in,
-                                              float*       out,
-                                              size_t       size);
+    typedef void (*InterleavingAudioCallback)(InterleavingInputBuffer in,
+                                              OutputBuffer            out,
+                                              size_t                  size);
 
     AudioHandle() : pimpl_(nullptr) {}
     ~AudioHandle() {}
--- a/src/hid/audio.h
+++ b/src/hid/audio.h
@@ -23,20 +23,38 @@ class AudioHandle
         ERR,
     };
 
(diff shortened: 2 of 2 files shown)


beserge:
Works totally fine on libdaisy as expected.
It does require changing all of the callbacks in DaisyExamples, which is no big deal.
My one concern is that this will break a lot of user's existing code.
It's a simple fix, but some people will be confused of course. Maybe we should put some messages out letting people know before pushing this.

stephenhensley:
Womp. This was already addressed in #330.

The required changes to DaisyExamples are why it hasn't gotten merged yet.

But I agree, we should make a general announcement before merging it because it does require changing some code. I think adding a typedef for the I/O buffers would make this a bit easier for people to understand anyway (since `float**` is also a fairly confusing symbol to folks newer to c/c++).

TheSlowGrowth:
This seems a little more complete than my version :-)

src/hid/audio.h
```diff
@@ -23,15 +23,22 @@ class AudioHandle
         ERR,
     };
 
+    typedef const float* const* InputBuffer;
+    typedef float*              OutputBuffer;
```
TheSlowGrowth: Why do you use a `float *` for both the interleaved and noninterleaved calbacks? The non-interleaved callback would need a `float **` or maybe even a `float * const *` (that is a const pointer to a non-const pointer if I'm not mistaken).

src/hid/audio.cpp
```diff
@@ -387,7 +387,7 @@ void AudioHandle::Impl::InternalCallback(int32_t* in, int32_t* out, size_t size)
                 break;
             default: break;
         }
-        cb((const float**)fin, (float**)fout, size / 2);
+        cb((InputBuffer)fin, (OutputBuffer)fout, size / 2);
```
TheSlowGrowth, line 390: Is this cast really necessary?

src/hid/audio.h
```diff
@@ -23,15 +23,22 @@ class AudioHandle
...
 
     /** Non-Interleaving Callback format. 
      ** audio is prepared as { L0, R0, L1, R1, . . . LN, RN }
      */
-    typedef void (*InterleavingAudioCallback)(const float* in,
-                                              float*       out,
-                                              size_t       size);
+    typedef void (*InterleavingAudioCallback)(InterleavingInputBuffer in,
+                                              OutputBuffer            out,
+                                              size_t                  size);
```
TheSlowGrowth: is this a total size or a size per channel? We could change the name to make it clearer

TheSlowGrowth (changes requested):

beserge:
DaisyExamples now has all examples modified with this format.
https://github.com/electro-smith/DaisyExamples/pull/192

src/hid/audio.h
beserge: This should be set up correctly... InputBuffer is for the Non-Interleaving, InterleavingInputBuffer is for the interleaving

src/hid/audio.h
beserge: It's the blocksize. So size / channel.

TheSlowGrowth: Yes I know, but it's not apparent from the current name, I think.

src/hid/audio.h
TheSlowGrowth: Ah I see that you chanegd it. Cool! Looking good!

src/hid/audio.cpp
beserge: So the way we have it now fin is declared as `fin float[size]`, which if I'm not mistaken allocates the space at compile time (maybe not since it depends on size to know how much space...). I'm not sure of a way to do this with our typedef without doing something like `InputBuffer fin = malloc(size)`

TheSlowGrowth:
```cpp
// The type of fin and fout is essentially float**, except that they're 
// both stack-allocated and thus local to the function.
float* fin[chns];
float* fout[chns];
// This should compile without problems
cb(fin, fout, size / 2);
```

stephenhensley: I agree, I think the cast is unnecessary here. (I had a redundant cast for `float**` in the previous code. So that's probably why it's here).

src/hid/audio.h
```diff
@@ -23,15 +23,23 @@ class AudioHandle
...
 
     /** Non-Interleaving Callback format. 
      ** audio is prepared as { L0, R0, L1, R1, . . . LN, RN }
      */
-    typedef void (*InterleavingAudioCallback)(float* in,
-                                              float* out,
-                                              size_t size);
+    typedef void (*InterleavingAudioCallback)(InterleavingInputBuffer  in,
+                                              InterleavingOutputBuffer out,
+                                              size_t                   size);
```
stephenhensley, line 42: I suggest we move the format documentation to the buffers themselves, and then add a separate bit of documentation for the Callbacks (that doesn't include the format information).

That'll help make it a bit easier to see what's what with editors using intellisense, and when browsing the generated documentation.

stephenhensley (changes requested):
Aside from the few notes I think this is solid.

stephenhensley (approved):

stephenhensley:
confirmed the DaisyExamples PR addressed the change as expected, and that things still make sound. Merging.

PR #345, opened 2021-05-17
Add conversion funcs for 8 bit ints to daisy_core

### LINKED ISSUE #305: Add conversion s28f() (int8_t to float)
Feature suggestion: it would be nice to have conversion int8_t to float to the daisy core.

https://github.com/electro-smith/libDaisy/blob/master/src/daisy_core.h

I used this today to use some 8bit SID noise code with DaisySP, that generates int8_t as its output.

```
#define S82F_SCALE 0.00390625f     /**< 1 / (2**8) */
#define F2S8_SCALE 255.0f          /**< (2 ** 8) - 1 */


FORCE_INLINE float s82f(int8_t x)
{
    return (float)x * S82F_SCALE;
}

FORCE_INLINE int8_t f2s8(float x)
{
    x = x <= FBIPMIN ? FBIPMIN : x;
    x = x >= FBIPMAX ? FBIPMAX : x;
    return (int32_t)(x * F2S8_SCALE);
}
```

beserge:
Fixes #305

### DIFF
--- a/src/daisy_core.h
+++ b/src/daisy_core.h
@@ -32,8 +32,8 @@ cache enabled.
 
 #define FBIPMAX 0.999985f             /**< close to 1.0f-LSB at 16 bit */
 #define FBIPMIN (-FBIPMAX)            /**< - (1 - LSB) */
-#define S82F_SCALE 0.00390625f        /**< 1 / (2**8) */
-#define F2S8_SCALE 255.0f             /**< (2 ** 8) - 1 */
+#define S82F_SCALE 0.0078125f        /**< 1 / (2**7) */
+#define F2S8_SCALE 127.0f             /**< (2 ** 7) - 1 */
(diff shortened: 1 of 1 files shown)


stephenhensley:
The scalar amount for the conversions should be based on 2^7 instead of 2^8 for signed 8-bit conversion. If this were casted to `int8_t` it would overflow at values where the absolute value is greater than 0.5

Worth noting, too. From what I've seen 8-bit PCM audio is generally unsigned.

stephenhensley:
The 8-bit unsigned conversions should include the offset (using the scale macros where applicable instead of inline division, and clamping the float input):
```c
FORCE_INLINE float u82f(uint8_t in)
{
    return ((float)in - 127) * (1.f / 127.f);
}
FORCE_INLINE uint8_t f2u8(float in)
{
  return (uint8_t)((in * 127.f) + 127.f);
}
```
Otherwise it'll overflow the uint8_t for negative float inputs with the current implementation.

PR #347, opened 2021-05-18
Spi update

beserge:
Updates the SPI peripheral to use the pimpl format.
Also adds configuration over
- peripheral
- mode
- direction
- clock polarity
- clock phase
- nss
- baudrate prescaling
- pins

The most pressing TODO is probably implementing functions to handle RX, since we have RX configs now, but no actual functions to let you RX.

### DIFF
--- a/src/per/spi.cpp
+++ b/src/per/spi.cpp
@@ -365,6 +365,9 @@ SpiHandle::Result SpiHandle::Impl::InitPins()
               && config_.direction == Config::Direction::TWO_LINES_RX_ONLY)
           || (!is_master && config_.direction == Config::Direction::TWO_LINES);
 
+    // nss = soft -> ss pin is unused for master and slave
+    bool enable_ss = config_.nss != Config::NSS::SOFT;
+
     if(config_.pin_config.sclk.port != DSY_GPIOX)
     {
         //check sclk against periph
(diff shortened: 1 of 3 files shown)


src/per/spi.cpp
```diff
@@ -8,126 +11,547 @@
...
+    if(HAL_SPI_Transmit(&hspi_, buff, size, 100) != HAL_OK)
+    {
+        return SpiHandle::Result::ERR;
+    }
+    return SpiHandle::Result::OK;
 }
 
-void HAL_SPI_MspInit(SPI_HandleTypeDef* spiHandle)
+SpiHandle::Result
+SpiHandle::Impl::PollReceive(uint8_t* buffer, uint16_t size, uint32_t timeout)
```
TheSlowGrowth, line 210: Seems like the corresponding function in the header file is mising

beserge: yep, fixed

stephenhensley:
Before I do the physical test for this, I figured I'd just leave a few notes on stuff I noticed looking through the files:

* The TODO at the top of the file can be updated. We should be able to get rid of "add configuration".
* The old `SpiPin` enum is [still in place](https://github.com/electro-smith/libDaisy/blob/76f44f6df9c16be4b5435b8c553f5dd6652dcba4/src/per/spi.h#L21-L28). It doesn't seem to be used anymore, I think we can get rid of it.
* `SpiHandle::Config::BaudPrescalar::BAUDRATEPRESCALER_2` is really verbose. I think we can truncate the member names to `PS_2`, `PS_4`, without it being any less clear.
    * Related, at some point I'd like to have a "target_frequency" member a real value in Hz to use instead of an arbitrary prescalar.
* I don't like the inconsistent naming between `BlockingTransmit` and `PollReceive`. For now lets make it consistent with the previously existing function. So `PollReceive`->`BlockingReceive`. And then post v0.1, we should go through and make all of the core peripherals have consistent naming for common function types.
* Related, we should update the BlockingTransmit function to have an (optional) timeout argument -- default can be 100ms since that's what's hard coded now.

Otherwise, the implementation looks solid. I'd probably mark all the `pin_alt_spi` stuff as static so that its clear that it should only ever be used in this file.

stephenhensley:
Cool. Thanks for making the changes. Still passes the tests I ran yesterday. I think it's good to go.

PR #348, opened 2021-05-19
Replace hardcoded pin deinit with function

### LINKED ISSUE #346: UartHandle::Impl::DeInitPins is unused
I defined this function to deinit the pins, but later in `HalMspDeinit `I rewrote the code rather than use the function.
Easy fix, just replace the repeated code with a function call.

beserge:
Hard coded in the deinit of the tx and rx pins for uart despite having written a function to do so.
This replaces that hardcoding with the function call.

Fixes #346

### DIFF
--- a/src/per/uart.cpp
+++ b/src/per/uart.cpp
@@ -609,13 +609,7 @@ void HAL_UART_MspDeInit(UART_HandleTypeDef* uartHandle)
             break;
     }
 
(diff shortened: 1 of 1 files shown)


stephenhensley:
Good catch!

beserge:
[libDaisy - GPIO C++ update](https://trello.com/c/26X32nTd/97-libdaisy-gpio-c-update)

PR #350, opened 2021-05-21
Midi handler wrapup

beserge:
So far, implements all 7 of the Channel Voice Messages tested working. 
TODO System Real-Time Messages and System Common Messages, Channel Mode Messages.


TheSlowGrowth:
Hi @beserge ! This is looking good! I wondered if we could add some unit tests for the MIDI parsing. It seems like a great chance to check for bugs, not just now but also in the future when other people want to make modifications to the MIDI classes. Ping me on slack if I can be of any help on this

stephenhensley:
Quick look over everything looks good. 
I agree with @TheSlowGrowth on adding unit tests to this. 

We'll have to add some preprocessor considerations to avoid the UART handle being a part of the class, but otherwise in the "NONE" input/output modes you could feed the Parse function with bytes in a test environment without any hardware reliance.

A few other things related that may be worth changing (either now or in a post v0.1 update):
* We could consider making the various input/output modes "Transports" like we did for the OLED. To keep the actual "MIDI parsing" separate from the hardware acquisition/transmission. That would help make this a bit more unit test compatible, and possibly easier to add the other modes down the road.
* @beserge and I were talking a bit earlier about possibly reworking `MidiEvent` struct, and the actual handling of MIDI Messages, etc. since the previously "simple" `MidiEvent` is pretty huge now that we've got pretty much full MIDI 1.0 support. I'm pretty split on a few different approaches. So I'm open to ideas and feedback.

In the interesting of limiting API breaking changes until we start having releases/changelogs to help people migrate I'm leaning toward making both of those points issues and handling them at a later time. With that said, idk if it would be better to hold off on unit tests until that happens or not.

TheSlowGrowth:
> To keep the actual "MIDI parsing" separate from the hardware acquisition/transmission.

> I suppose the Parse(byte) functionality isn't likely to change too much. So might not be a bad thing to add tests sooner than later.

100% agree to that. We've seen on the OledDisplay stuff that the transport abstraction doesn't have to come with breaking API changes. It would allow to add tests right now and ship a very solid implementation for v0.1.

stephenhensley:
>We've seen on the OledDisplay stuff that the transport abstraction doesn't have to come with breaking API changes.

It did require some changes to user code for the Initialization API, though. But I agree that it's worth making this change sooner than later regardless.

So yeah, @beserge I think we should make the following changes:
* Move the MIDI Event stuff (as-is) to it's own `MidiEvent.h` file. (that'll help keep the file cleaner with the next few changes)
* Add `MidiUartTransport` class that can be used for the next point.
* Make the `MidiHandler` a template class with a Transport argument as illustrated below
    * this will remove the need for the current init arguments
    * a Config struct (containing the transport config struct) should be provided for configuring UART pins, and any other related config we want to set up later (I've seen some MIDI libraries have the ability to just ignore certain message types which might be a nice feature to add down the road).
* Add UnitTests for the new non-hardware-reliant MidiHandler
    * I think @TheSlowGrowth will have some good advice for this
    * We'll probably need to make a "TestTransport" class or something, but that can probably live in the test itself.

I think that'll pretty much cover it for now.

stephenhensley:
Also, if we want to hide away the templating a bit for users we can pre-typedef a class like we did for the various common display types:

```
using MidiUartHandler = MidiHandler<MidiUartTransport>;
```

beserge:
OK moved to the Transport method.
Wrote a UartTransport and a TestTransport. The test is just a ton of hardcoded test data. 
Will move to a method that just loops over all possible (valid) messages for the unit test.
Also: seriously dropped the size of the sysex buffer. It was huge and was causing overflows when I tried to build.

stephenhensley:
this works, and fixes some issues that exist in the current implementation (i.e. Clock messages stalling the state machine, etc.).

I do want to add unit tests to validate the parser functionality before we make further changes, or rework any of the MidiEvent stuff, but I don't want to leave MIDI partially broken in the meantime. So I think for now this is good to go.

PR #352, opened 2021-05-27
Feature/CPU load meter

TheSlowGrowth:
This PR adds a CPU load meter to measure the CPU load of your audio callbacks.

### DIFF
--- a/src/daisy.h
+++ b/src/daisy.h
@@ -44,6 +44,7 @@
 #include "dev/codec_wm8731.h"
 #include "dev/lcd_hd44780.h"
 #include "util/scopedirqblocker.h"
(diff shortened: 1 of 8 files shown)


stephenhensley (approved):

stephenhensley:
Looks good! And yet again, the unit tests made this super easy to review.

Something we may want to add in the future if we want a cycle-resolution is use the DWT registers for the clock cycle counter.

It was mentioned a while ago in the forum on [this thread](https://forum.electro-smith.com/t/solved-how-to-do-mcu-utilization-measurements/1236/3). Obviously, it'll be a platform dependent bit, but might be useful to have that super high resolution timing for something.

PR #353, opened 2021-05-28
Spi Init NSS bugfix

beserge:
Nss switch case was setting clock_phase not nss.
Need to be more careful copy/pasting in the future.


stephenhensley:
Approved

PR #354, opened 2021-06-03
CPPify GPIO and Pin

beserge:


beserge:
[libDaisy - GPIO C++ update](https://trello.com/c/26X32nTd/97-libdaisy-gpio-c-update)

beserge:
OK this is working well from the limited tests I've made. I'm _certain_ there are bugs in here, but it's a big PR, so it would really benefit from the better unit test coverage we have coming down the pike.

Other than that, some reviews on the `GPIO `and `Pin` classes themselves would be good. I also updated `hal_map` to cpp as part of this.

src/per/adc.cpp
```diff
@@ -153,30 +93,31 @@ static int get_num_mux_pins_required(int num_mux_ch)
...
-        PIN_CHN_14,
-        PIN_CHN_15,
-        PIN_CHN_16,
-        PIN_CHN_17,
-        PIN_CHN_18,
-        PIN_CHN_19,
-    };
+    Pin adcpins[DSY_ADC_MAX_CHANNELS];
+
+    adcpins[0].Init(Pin::Port::DSY_GPIOA, 6);
```
TheSlowGrowth, line 103: use `{Pin::Port::DSY_GPIOA, 6}` notation after `Pin` constructor is removed

src/per/adc.cpp
```diff
@@ -353,22 +291,22 @@ void AdcHandle::Init(AdcChannelConfig* cfg,
     sConfig.Offset       = 0;
     for(uint8_t i = 0; i < adc.channels; i++)
     {
-        const auto& cfg = adc.pin_cfg[i];
+        auto& cfg = adc.pin_cfg[i];
```
TheSlowGrowth, line 294: why remove the const here?

src/per/adc.h
```diff
@@ -48,15 +48,17 @@ struct AdcChannelConfig
...
-                 dsy_gpio_pin mux_1 = {DSY_GPIOX, 0},
-                 dsy_gpio_pin mux_2 = {DSY_GPIOX, 0});
-
-    dsy_gpio pin_;                   /**< & */
-    dsy_gpio mux_pin_[MUX_SEL_LAST]; /**< & */
-    uint8_t  mux_channels_;          /**< & */
+    void InitMux(Pin    adc_pin,
+                 size_t mux_channels,
+                 Pin    mux_0,
+                 Pin    mux_1 = Pin(),
```
TheSlowGrowth, line 54: We could also provide `Pin::invalid` like this
```cpp
struct Pin
{
    // ...
    static constexpr Pin invalid = {DSY_GPIOX, 0};
}
```
and write:
```cpp
void InitMux(Pin    adc_pin,
                 size_t mux_channels,
                 Pin    mux_0,
                 Pin    mux_1 = Pin::invalid,
                 Pin    mux_2 = Pin::invalid);
```

src/per/dac.cpp
```diff
@@ -238,19 +235,20 @@ void DacHandle::Impl::InitPins()
     if(ChannelOneActive())
     {
         // Init PA4 as Analog
-        dsy_gpio d1;
-        d1.pin  = {DSY_GPIOA, 4};
-        d1.mode = DSY_GPIO_MODE_ANALOG;
-        dsy_gpio_init(&d1);
+        GPIO::Config gpio_conf;
+        gpio_conf.pin.Init(Pin::Port::DSY_GPIOA, 4);
```
TheSlowGrowth, line 239: use `{...}` after `Pin` constrcutor is removed

src/per/dac.cpp
```diff
@@ -238,19 +235,20 @@ void DacHandle::Impl::InitPins()
...
     if(ChannelTwoActive())
     {
         // Init PA5 as Analog
-        dsy_gpio d2;
-        d2.pin  = {DSY_GPIOA, 5};
-        d2.mode = DSY_GPIO_MODE_ANALOG;
-        dsy_gpio_init(&d2);
+        GPIO::Config gpio_conf;
+
+        gpio_conf.pin.Init(Pin::Port::DSY_GPIOA, 5);
```
TheSlowGrowth, line 249: use `{}` after `Pin` constrcutor is removed

TheSlowGrowth (changes requested):
Haven't yet looked at the QSPI

stephenhensley:
I agree with @TheSlowGrowth's notes so far. I'll have some time later this week to give this a review myself.

src/per/adc.cpp
beserge: We need to do `cfg.pin.Init()` for all of those GPIO in cfg, which means we can't use const.

src/per/adc.h
beserge: Are you sure this is compatible with Pin being POD?
I get `a member of type "const daisy::Pin" cannot have an in-class initializer`
I'm sure it'll work if we have INVALID_PIN as a define, or maybe just not a member of Pin.

src/daisy_seed.h
```diff
@@ -134,14 +134,16 @@ class DaisySeed
...
-    dsy_sdram_handle sdram_handle; /**< & */
-    dsy_qspi_handle  qspi_handle;  /**< & */
-    AudioHandle      audio_handle; /**< & */
-    AdcHandle        adc;          /**< & */
-    DacHandle        dac;
-    UsbHandle        usb_handle; /**< & */
-    dsy_gpio         led, testpoint;
-    System           system;
+    QSPIHandle          qspi;
+    QSPIHandle::Config  qspi_config;
```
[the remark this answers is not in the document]
beserge: I'm fine with this. It does mean `Seed::Config` will now be empty, so we'll have to adjust some stuff in DaisyExamples.

src/per/gpio.h
```diff
@@ -2,75 +2,125 @@
...
+        DSY_GPIOK, /**< & */
+        DSY_GPIOX, /** This is a non-existant port for unsupported bits of hardware. */
+        DSY_GPIO_LAST, /** Final enum member */
+    };
 
-    /** @addtogroup other
-    @{
-    */
+    //allow equality operator to be used, replaces pin_cmp
+    bool operator==(Pin &rhs) { return (rhs.pin == pin) && (rhs.port == port); }
```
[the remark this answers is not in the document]
beserge: Default comparison operator doesn't work unfortunately.
`error: no match for 'operator==' (operand types are 'daisy::Pin' and 'daisy::Pin')`

src/per/spi.cpp
```diff
@@ -341,19 +335,14 @@ SpiHandle::Result checkPinMatchSpi(GPIO_InitTypeDef* init, Pin pin, int p_num)
...
     {
-        Pin tmp_pin;
-        tmp_pin.Init(pins_periphs_spi[p_num][i].port,
-                     pins_periphs_spi[p_num][i].pin);
-
         Pin none_pin;
-        none_pin.Init(pins_none_spi.port, pins_none_spi.pin);
 
-        if(tmp_pin == none_pin)
+        if(pins_periphs_spi[p_num][i].pin == none_pin)
```
TheSlowGrowth, line 340: Have you tried to implement `Pin::invalid`? It would make this comparison a little prettier by removing the need for `none_pin`
```cpp
struct Pin 
{
    // ...
    static constexpr Pin invalid = { Port::DSY_GPIOX, 0 };
}
```

src/per/uart.cpp
```diff
@@ -342,18 +341,15 @@ UartHandler::Result checkPinMatch(GPIO::Config init, Pin pin, int p_num)
...
 
         // Pin() defaults to GPIOX, 0
-        if(temp_pin == empty_pin)
+        if(pins_periphs[p_num][i].pin == empty_pin)
         {
             /* skip */
         }
 
-        else if(temp_pin == pin)
+        else if(pins_periphs[p_num][i].pin == pin)
```
TheSlowGrowth, line 352: How about this:
```cpp
if (pins_periphs[p_num][i].pin.IsValid())
   // ...
   
struct Pin 
{
    // ...
    bool IsValid() const { return port != Port::DSY_GPIOX; }
}
```

TheSlowGrowth (changes requested):

src/per/adc.h
TheSlowGrowth: Hmm, you're right, that's weird.
[This solution works though](https://godbolt.org/#g:!((g:!((g:!((g:!((h:codeEditor,i:(fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:20,positionColumn:2,positionLineNumber:20,selectionStartColumn:2,selectionStartLineNumber:20,startColumn:2,startLineNumber:20),source:'%23include+%3Cstdint.h%3E%0A%0Aenum+class+Port%0A%7B%0A++++DSY_GPIOA,%0A++++//+...%0A++++DSY_GPIOX%0A%7D%3B%0A%0Astruct+Pin%0A%7B%0A++++Port+port+%3D+Port::DSY_GPIOX%3B%0A++++uint8_t+pin+%3D+0%3B%0A++++static+constexpr+Pin+invalid()+%7B+return+%7B+Port::DSY_GPIOX,+0+%7D%3B+%7D%0A%7D%3B%0A%0Avoid+test()%0A%7B%0A++++Pin+p+%3D+Pin::invalid()%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:47.74396642182581,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g111,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'x86-64+gcc+11.1+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:52.256033578174176,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',m:63.31811263318112,n:'0',o:'',t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,fontUsePx:'0',wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+11.1+(Compiler+%231)',t:'0')),header:(),l:'4',m:36.68188736681888,n:'0',o:'',s:0,t:'0')),l:'3',n:'0',o:'',t:'0')),version:4)

TheSlowGrowth (approved):

stephenhensley:
I've marked this as a draft for now, and renamed to a more fitting name given that #451 handles the new types. 

It might be worth just using this as a reference, and doing individual PRs to update each relevant peripheral/device to use the new `Pin`/`GPIO` structs in stead of the old types. It'll definitely be easier to review.

Fortunately, for the time being the new `Pin` object can be dropped in anywhere that the `dsy_gpio_pin` was previously used.

stephenhensley:
Due to the massive amount of conflicts, and the fact that it might be cleaner to piece-meal update the affected internal bits, Im going to close this for the time being

the `Pin` and `GPIO` have been added in master, and are usable with all typical user-facing APIs

PR #359, opened 2021-06-16
SPI DMA implementation

brbrr:
This is the first config-less iteration on adding DMA support to SPI. This configuration is battle-tested with ILI9341-based 240x320 TFT display. 

I'm opening this PR mostly to gather some feedback on whether it is needed at all and get suggestions on the best way to implement it.

Obviously, DMA init should be configurable similar to SPI. If this PR would make sense, I'll work on adding configuration support as a next step.

### DIFF
--- a/src/per/spi.cpp
+++ b/src/per/spi.cpp
@@ -814,7 +814,7 @@ SpiHandle::Result SpiHandle::Impl::DeInitPins()
 
 volatile int8_t SpiHandle::Impl::dma_active_peripheral_;
 SpiHandle::Impl::SpiDmaJob
-    SpiHandle::Impl::queued_dma_transfers_[kNumSpiWithDma];
+                               SpiHandle::Impl::queued_dma_transfers_[kNumSpiWithDma];
 SpiHandle::CallbackFunctionPtr SpiHandle::Impl::next_callback_;
 void*                          SpiHandle::Impl::next_callback_context_;
 
--- a/src/per/spi.cpp
+++ b/src/per/spi.cpp
@@ -25,19 +25,13 @@ class SpiHandle::Impl
     uint8_t dma_flag = 0;
 
   public:
-    enum
-    {
-        DMA_RX,
-        DMA_TX,
-    };
-
     struct SpiDmaJob
     {
         uint8_t*                       data             = nullptr;
         uint16_t                       size             = 0;
         SpiHandle::CallbackFunctionPtr callback         = nullptr;
         void*                          callback_context = nullptr;
-        uint8_t                        direction        = DMA_TX;
+        SpiHandle::DmaDirection        direction = SpiHandle::DmaDirection::TX;
 
         bool IsValidJob() const { return data != nullptr; }
         void Invalidate() { data = nullptr; }
@@ -74,7 +68,7 @@ class SpiHandle::Impl
     static bool IsDmaTransferQueuedFor(size_t spi_idx);
 
     Result SetDmaPeripheral();
-    Result InitDma(uint8_t direction);
+    Result InitDma(SpiHandle::DmaDirection direction);
 
     uint8_t GetDmaFlag() { return dma_flag; }
     void    SetDmaFlag(uint8_t f) { dma_flag = f; }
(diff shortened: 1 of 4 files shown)


TheSlowGrowth:
I quickly looked into the code. 
- I'm not sure about the DmaFlag() functions, they look like an exposed implementation detail to me.
- Multiple SpiHandles must be able to start DMA transmissions simultaneously without interfering with each other. Right now, you can effectively only use DMA for a single SpiHandle without everything breaking apart.

I think regarding the overall structure, you can take a look at the I2C implementation. There we have a "scheduling" system so that multiple I2CHandles can share a single DMA which will transmit scheduled jobs on these individual handles one after the other. Without this scheduling, you end up with collisions between multiple handles that want to transmit with DMA simultaneously.

brbrr:
Thanks, @TheSlowGrowth! I followed your suggestion and implemented scheduling similarly (ctrl-c/ctrl-v mostly 😬) to I2C implementation. 

Unfortunately, I running into some sort of a race condition when calling `DmaTransmit` multiple(10) times in a loop. I can't really figure out what's going on, but here is what I discovered so far:

- Only 2 calls out of 10 is being processed, specifically, only first and last one.
- When libDaisy optimizations disabled (`-O3`) all the calls processed in order

src/per/spi.cpp
```diff
@@ -693,21 +870,39 @@ extern "C" void SPI1_IRQHandler(void)
...
+    ScopedIrqBlocker block;
+    if(SpiHandle::Impl::dma_active_peripheral_ >= 0)
+        HAL_DMA_IRQHandler(
+            &spi_handles[SpiHandle::Impl::dma_active_peripheral_].hdma_spi_);
+}
 extern "C" void DMA1_Stream7_IRQHandler(void)
 {
-    HAL_DMA_IRQHandler(&spi_handles[0].hdma_spi_);
+    HalSpiDmaStreamCallback();
+    // HAL_DMA_IRQHandler(&spi_handles[0].hdma_spi_);
```
TheSlowGrowth, line 886: remember to remove this when you're ready ;-)

src/per/spi.cpp
```diff
@@ -693,21 +870,39 @@ extern "C" void SPI1_IRQHandler(void)
...
 }
 
-
 extern "C" void HAL_SPI_TxCpltCallback(SPI_HandleTypeDef* hspi)
 {
-    SpiHandle::Impl* handle = MapInstanceToHandle(hspi->Instance);
+    SpiHandle::Impl::DmaTransferFinished(hspi, SpiHandle::Result::OK);
+    // SpiHandle::Impl* handle = MapInstanceToHandle(hspi->Instance);
```
TheSlowGrowth, line 892: remember to remove this when you're ready ;-)

src/per/spi.cpp
```diff
@@ -743,11 +938,14 @@ SpiHandle::BlockingReceive(uint8_t* buffer, uint16_t size, uint32_t timeout)
     return pimpl_->BlockingReceive(buffer, size, timeout);
 }
 
-SpiHandle::Result SpiHandle::DmaTransmit(uint8_t* buff, size_t size)
+SpiHandle::Result
+SpiHandle::DmaTransmit(uint8_t*                       buff,
```
TheSlowGrowth, line 945: You haven't yet added the queue to the DmaReceive function

TheSlowGrowth (changes requested):
Here's a couple of things I noticed. I'll be taking a closer look later when I have more time

src/per/spi.cpp
```diff
@@ -22,38 +25,73 @@ class SpiHandle::Impl
     uint8_t dma_flag = 0;
 
   public:
+    enum
```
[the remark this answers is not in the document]
brbrr: I think you reviewed outdated revision. I have already done it and moved it into a header file: 
b5bcbb5

brbrr:
Hey @TheSlowGrowth. I think I addressed all of your feedback

TheSlowGrowth (commented):
Good job it's looking very nice and I don't see anything obvious. I'll try to get this running on some hardware later to see it in action.

brbrr:
FWIW I was not been able to reproduce the styling issue locally with VSCode and via `clang-fromat` executable installed from brew.

stephenhensley:
Overall looking very good! Will do some actual testing on the hardware this week, and should be able to get it merged if there are no issues!

stephenhensley:
@brbrr apologies on the delay here, but we finally found some time to do a bit of hardware testing, and it works!

I think I'm signing off on everything here!

We'll take care of the addition to the changelog for this, but in the future if you could add an entry to the changelog along with your PR that would be great.

Thanks again so much for the contribution!

PR #360, opened 2021-06-23
Midi unit test

### LINKED ISSUE #358: Add unit tests for MIDI parser
We should be able to do most of the testing just feeding bytes into the Parse function. However, having a dummy transport that can just take a predefined array to feed in, would help test the queue system as well.

We should validate:

* the state at Initalization 
* the reception of all standard MIDI types
* overflow conditions
* invalid message handling
* filtering of specified note types works (this would require that #357 already happened, but can be added later if we want to add tests first.) 

There may be a few things I missed, but that should be pretty good coverage.

beserge:
Fixes #358

### DIFF
--- a/tests/Midi_gtest.cpp
+++ b/tests/Midi_gtest.cpp
@@ -198,21 +198,20 @@ TEST_F(MidiTest, systemRealTime)
 // sysex
 TEST_F(MidiTest, systemExclusive)
 {
-    uint8_t msgs[110];
-    for(int i = 0; i < 110; i++){
+    uint8_t msgs[135];
+    for(int i = 0; i < 135; i++){
         msgs[i] = (uint8_t)i;
     }
 
     ParseAndPopSysex(msgs, 6);
     TestSysex(msgs, 6);
 
-    ParseAndPopSysex(msgs, 100);
-    TestSysex(msgs, 100);
+    ParseAndPopSysex(msgs, 128);
+    TestSysex(msgs, 128);
 
-    //max len is 100, let's go past that
-    //the default behavior is to just cut the message off there, maybe that should be changed
-    ParseAndPopSysex(msgs, 110);
-    TestSysex(msgs,100);
+    //max len is 128, let's go past that
+    ParseAndPopSysex(msgs, 135);
+    TestSysex(msgs,128);
 
     //queue should be empty
     while(midi.HasEvents()){
--- a/tests/Midi_gtest.cpp
+++ b/tests/Midi_gtest.cpp
@@ -272,30 +272,141 @@ TEST_F(MidiTest, pitchBend)
     EXPECT_FALSE(midi.HasEvents());
 }
 
-//Channel Mode Messages
-// also doesn't totally test channels (for now)
-TEST_F(MidiTest, channelMode)
+// ================ Channel Mode Messages ================ 
+
+TEST_F(MidiTest, allSoundOff)
 {
-    //All messages
-    for(uint8_t type = 120; type < 128; type++)
+    for(uint8_t chn = 0; chn < 16; chn++)
     {
-        uint8_t msg[] = {0x80 + (3 << 4), type, 0};
+        uint8_t msg[] = {(uint8_t)(0x80 + (3 << 4) + chn), 120, 0};
         MidiEvent event = ParseAndPop(msg, 3);
-        EXPECT_EQ((uint8_t)event.cm_type, type - 120);
+        AllSoundOffEvent asoEvent = event.AsAllSoundOff();
+
+        EXPECT_EQ((uint8_t)event.type, ChannelMode);
+        EXPECT_EQ((uint8_t)event.cm_type, AllSoundOff);
+        EXPECT_EQ((uint8_t)asoEvent.channel, chn);
     }
 
-    //LocalControlOn
-    uint8_t msg[] = {0x80 + (3 << 4), 122, 127};
-    MidiEvent event = ParseAndPop(msg, 3);
-    Test(event, (uint8_t)event.cm_type, (uint8_t)LocalControl, 0, true, 122, 127);
+    EXPECT_FALSE(midi.HasEvents());
+}
 
-    //MonoModeOn
-    for(uint8_t data = 0; data < 128; data++)
+TEST_F(MidiTest, resetAllControllers)
+{
+    for(uint8_t chn = 0; chn < 16; chn++)
     {
-        uint8_t msg[] = {0x80 + (3 << 4), 126, data};
-        MidiEvent event = ParseAndPop(msg, 3);
-        Test(event,(uint8_t)event.cm_type, (uint8_t)MonoModeOn, 0, true, 126, data);
+        for(uint8_t val = 0; val < 128; val++){
+            uint8_t msg[] = {(uint8_t)(0x80 + (3 << 4) + chn), 121, val};
+            MidiEvent event = ParseAndPop(msg, 3);
+            ResetAllControllersEvent racEvent = event.AsResetAllControllers();
+
+            EXPECT_EQ((uint8_t)event.type, ChannelMode);
+            EXPECT_EQ((uint8_t)event.cm_type, ResetAllControllers);
+            EXPECT_EQ((uint8_t)racEvent.channel, chn);
+            EXPECT_EQ((uint8_t)racEvent.value, val);
+        }
     }
+
+    EXPECT_FALSE(midi.HasEvents());
+}
+
+TEST_F(MidiTest, localControl)
+{
+    //Off
+    for(uint8_t chn = 0; chn < 16; chn++)
+    {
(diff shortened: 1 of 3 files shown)


tests/Midi_gtest.cpp
```diff
@@ -0,0 +1,46 @@
...
+    uint8_t Rx() { return 1; }    
+    bool RxActive() { return true; }
+
+    private:
+};
+
+TEST(util_MidiHandler, a_stateAfterInit){
+    MidiHandler<MidiTestTransport> midi;
+    MidiHandler<MidiTestTransport>::Config handler_conf;
+    midi.Init(handler_conf);
```
TheSlowGrowth, line 33: I suspect this initialization will be needed for every test. You could consider to make a test fixture that does this for you so you don't have to repeat this over and over

TheSlowGrowth: Alternatively you could consider parameterized tests, where the test parameters look like this:

```cpp
struct TestParameters
{
    std::string name; // human readable name of the test scenario for displaying in test logs
    std::array<uint8_t> inputStream; // raw input data
    std::array<MidiEvent> expectedEvents; // expected events
};
```

Then you would write a single test that pushes all bytes into the parser and then reads the results one by one, comparing them to the expected results.
You could then provide multiple instantiations of this test for various test cases, e.g. running status, super long sysex, broken/invalid messages, etc.

Not sure how familiar you are with the Googletest framework. It's not hard once you know the basic pattern. I'd be happy to help you out if you like, just let me know.

tests/Midi_gtest.cpp
```diff
@@ -0,0 +1,46 @@
+#include <gtest/gtest.h>
+#include "hid/midi.h"
```
TheSlowGrowth, line 2: You can include "system.h" to fix the build error. The `System::Now()` function is already available in a unit test compatible, mocked form

TheSlowGrowth:
@beserge Great to see you're adding these tests! It'll be amazing to have the parser fully tested! 🚀

stephenhensley:
So these are passing, and they look solid to me.

@TheSlowGrowth you're definitely more familiar with tests than I am. Anything in particular stand out to you?

The only thing that stood out to me that all of the channel-voice messages get tested in one big loop, which is probably fine, but I'm not sure if it'd be more typical for them to get split up a bit more or not.

tests/Midi_gtest.cpp
```diff
@@ -0,0 +1,288 @@
...
+                {
+                    uint8_t msgs[]
+                        = {(uint8_t)(0x80 + (type << 4) + chn), d0, d1};
+
+                    bool sendThree = type != 4 && type != 5;
+                    ParseAndPop(msgs, 2 + sendThree);
+
+                    uint8_t chkType
+                        = (type == 3 && d0 > 119) ? (uint8_t)ChannelMode : type;
+                    Test((uint8_t)event.type, chkType, chn, sendThree, d0, d1);
```
TheSlowGrowth, line 145: I would prefer to have the `EXPECT_EQ()` statements directly here in the test body instead of in the `Test()` function.

- makes it easier to understand what the test does when reading the test body. I know it seems logical to extract this into a shared function when it occurs in multiple tests. But as a reader, you'll have to scroll up and trace which variable from the test ends up in which `EXPECT_EQ` to understand what's going on. 
- when the test fails, the output will contain the line number of the `EXPECT_` macro that failed. If this line number is burried in a function somewhere, it's harder to trace back why it failed.

tests/Midi_gtest.cpp
```diff
@@ -0,0 +1,288 @@
...
+                for(uint8_t d1 = 0; d1 < 128; d1++)
+                {
+                    uint8_t msgs[]
+                        = {(uint8_t)(0x80 + (type << 4) + chn), d0, d1};
+
+                    bool sendThree = type != 4 && type != 5;
+                    ParseAndPop(msgs, 2 + sendThree);
+
+                    uint8_t chkType
+                        = (type == 3 && d0 > 119) ? (uint8_t)ChannelMode : type;
```
TheSlowGrowth, line 144: Please add a comment why this is needed

tests/Midi_gtest.cpp
```diff
@@ -0,0 +1,288 @@
...
+    for(uint8_t data = 0; data < 128; data++)
+    {
+        uint8_t msg[] = {0x80 + (3 << 4), 126, data};
+        ParseAndPop(msg, 3);
+        Test((uint8_t)event.cm_type, (uint8_t)MonoModeOn, 0, true, 126, data);
+    }
+}
+
+//System Common Messages (no sysex)
+TEST_F(MidiTest, systemCommon)
```
TheSlowGrowth, line 179: Dito here: I would prefer these as separate tests to also cover the `asXYZ()` functions

tests/Midi_gtest.cpp
```diff
@@ -0,0 +1,288 @@
...
+    for(uint8_t i = 0; i < 20; i++)
+    {
+        msgs[0] = msgs[1] = i;
+        ParseAndPop(msgs, 2);
+        Test((uint8_t)event.type, chkType, 3, true, i, i);
+    }
+}
+
+//Bad data
+TEST_F(MidiTest, badData)
```
TheSlowGrowth, line 270: Nice!

tests/Midi_gtest.cpp
```diff
@@ -0,0 +1,288 @@
...
+        msgs[i] = (uint8_t)i;
+    }
+
+    ParseAndPopSysex(msgs, 6);
+    TestSysex(msgs, 6);
+
+    ParseAndPopSysex(msgs, 128);
+    TestSysex(msgs, 128);
+
+    //max len is 128, let's go past that
```
TheSlowGrowth, line 230: Nice!

TheSlowGrowth (changes requested):

TheSlowGrowth:
After I made my comments yesterday, I kept thinking about how to strike the perfect balance between "avoiding code repetition" and "keeping the relevant assertions/checks inside the test body". I have a gut feeling of what I prefer but I find it quite hard to express it in words.

On the one hand, code repetition is bad: 
- It makes tests harder to understand ("What's the difference between test A and test B? They look identical!")
- It's a burden for changing the API if you have to fix the same problem in many, many of tests

On the other hand, extracting a lot of the test code into shared functions is also bad:
- It makes the test harder to understand ("What's this function doing?") and more prone to errors (e.g. mixed up order of function arguments)
- Error output from the test runner is less helpful (`EXPECT_EQ()` inside a `checkResult()` function => "Failed assertion in test.cpp:36" => "`checkResult()` is called multiple times in my test, which call was the problem?")
- It hides the intention of the test 

I thought a lot about how to balance these two sides and I think it comes down to this: 
A test is not just a verification of the behavior of some code, it's also a documentation of what the expected behavior is. For someone who wants to refactor something, this is critical. Ideally, the test should be as short as possible while being understandable for someone who doesn't even know the language.

For me that means that I prefer to remove as much of the "tooling" and housekeeping" from the test body as possible and push that into functions that have descriptive names, like the `ParseAndPop()` function in this PR. What remains in the test should ideally be almost like a spec sheet:
```
// do something
// expect some result
// do something else
// expect some other result
```
When you look at unit tests in the Javascript/web world, the tests look like sentences even for a non-programmer. Take this (modified) example from the `Mocha` testing framework documentation, that tests the `indexOf()` function of an `Array` class:
```javascript
describe('Array', () => {
  describe('#indexOf()', () => {
    it('should return the correct index', () => {
      [1, 2, 3].indexOf(1).should.equal(0);
      [1, 2, 3].indexOf(2).should.equal(1);
      [1, 2, 3].indexOf(3).should.equal(2);
    });
    it('should return -1 when the value is not present', () => {
      [1, 2, 3].indexOf(5).should.equal(-1);
      [1, 2, 3].indexOf(0).should.equal(-1);
    });
  });
});
```
You could print this as a documentation! I can look at this and immediately say "Oh, that's what it's meant to do!"

Coming back to the MIDI tests: I think that's why I would keep most of the `EXPECT_` inside the test body, to clearly express expectations. For cases where checking these expectations is very complex (e.g. verify a transfer function of a filter) I would push the checks into a shared function (e.g. `bool checkTransferFunctionMatches(measuredTransferFunction, referenceTransferFunction)`) and keep an `EXPECT_TRUE(checkTransferFunctionMatches(...));` in the test body. That way I could still clearly express the expectation while keeping the bulk of the code away from the test body where it would only interfere with the "sentence-like" structure.

These are my experiences and obviously YMMV. I'm writing this down so long-winded because I wish I had this information when I started with automated testing. In the end I learned SO MUCH from colleagues in code reviews. Maybe I can pass some of that knowledge on now :-)

In the end, any test is better than no test, so feel free to use or not use from this what you like. :-)

TheSlowGrowth (approved):

stephenhensley (approved):
Looks good. The tests are definitely longer, but its much easier to see that there is coverage over all the different events.

PR #361, opened 2021-06-23
updated qspi to cpp standard

CorvusPrudens:
The compiled size on the example I used increased by 536 bytes, from 44,028 to 44,564.

Like i2c, this module takes a config struct for configuration. It may not align with other classes in `daisy_seed.h`.

I didn't test the CMake build, but it should work with my single change to the cmakelists file.


CHANGELOG.md
```diff
@@ -0,0 +1,29 @@
+# libDaisy Changelog
+
+## Unreleased
+
+* Update TODO comment in uart.h to reflect most recent uart update.
+* Add MidiEvent and related data to daisy namespace.
+* Update qspi API from C to C++.
```
TheSlowGrowth, line 7: Both of the above lines should have a "BREAKING CHANGE: " prefix, IMHO

stephenhensley: Good call! Yeah, I figured it'd be obvious by the migration notes below, but as clear as we can be about that, the better.

CorvusPrudens:
[Update change log for PRs](https://trello.com/c/MKRYtbxE/148-update-change-log-for-prs)

stephenhensley (changes requested):
Just a few little changes.

Another I think we should do is rework the mode switching so that the `Erase` and `Write` functions change internally update the mode of operation without having to have the user concern themselves with it. This will simplify the usage a bit.

With that, we should probably move the `Config::Mode` out of the public Config and track the mode inside the `Impl` class with a public getter to basically see if the QSPI is in Write or Read mode. We can elaborate that in a later update when we want to add DMA or IRQ based writes so that we don't have  block execution when writing to the QSPI.

CorvusPrudens:
[C++-ify QSPI peripheral code](https://trello.com/c/hmQiSNzI/123-c-ify-qspi-peripheral-code)

src/per/qspi.cpp
```diff
@@ -0,0 +1,1008 @@
...
+
+    if(reset_mode)
+        SET_MODE(Mode::MEMORY_MAPPED);
+    return QSPIHandle::Result::OK;
+
+// I'm sorry, but it does minimize code size here
+write_error:
+    if(reset_mode)
+        SET_MODE(Mode::MEMORY_MAPPED);
+    return QSPIHandle::Result::ERR;
```
stephenhensley, line 247: how much code size, because I really don't like this.

even just doing:
```
status = writeenable();
if (status == OK)
  status = command()
if (status == OK)
  status = transmit()
. . . etc.
if (reset_mode)
  SetMode(Mode::MemoryMapped);
return status;
```

looks like it be as cheap, or just barely more instructions.

src/per/qspi.cpp
```diff
@@ -0,0 +1,1008 @@
...
+    if(mode_ != mode)                               \
+    {                                               \
+        mode_ = mode;                               \
+        if(Init(config_) != QSPIHandle::Result::OK) \
+        {                                           \
+            mode_ = Mode::MEMORY_MAPPED;            \
+            Init(config_);                          \
+            return QSPIHandle::Result::ERR;         \
+        }                                           \
+    }
```
stephenhensley, line 25: why is this a macro? This can/should be an Impl function.

Having it be a functions should also help with the notes on "code size" where you're using gotos and jumping to this macro. Without it being a function, every instance is forced into the flash size, and can't easily be optimized.

src/per/qspi.cpp
```diff
@@ -0,0 +1,1008 @@
...
+        goto erase_error;
+    }
+
+    SET_MODE(Mode::MEMORY_MAPPED);
+    return QSPIHandle::Result::OK;
+
+// I'm sorry, but it does minimize code size here
+erase_error:
+    SET_MODE(Mode::MEMORY_MAPPED);
+    return QSPIHandle::Result::ERR;
```
stephenhensley, line 407: see note on other goto statement.

stephenhensley (changes requested):
Only other thing is that we should add a check to the init function to make sure that the program isn't executing from the QSPI before taking action.

We should add a Status retval for this as well to make it clear to a user why it's failing if they try it, and read the retval.

stephenhensley:
I'll probably beef up the changelog entry a bit more before we do the next major release, but I've tested this quite a bit today and aside from what I can only rule as an out of sync library build of some kind, everything has been working great.

stephenhensley (approved):

PR #366, opened 2021-06-29
USB MIDI functionality

CorvusPrudens:
This adds USB MIDI support with an interface compatible with the `MidiHandler`. Sysex send is not supported, but all other messages in and out should be. It's possible the endpoints are not perfectly set up, since they're identical to the standard CubeMX CDC ones, but so far for me it seems to be working perfectly. We will likely want to expand or modify the descriptor strings for MIDI mode so it's easier to identify in a DAW.


h-iggs:
That is sooo cool :) And it works too. Just tweaked my daisy_field.cpp and tada! USB MIDI. Thank you :)


Clearly I'm not a PR approver just a grateful user.

TheSlowGrowth:
I haven't actually found time to look at this or test this, but ... Does it support multiple virtual midi ports? I know the midibox project does. Maybe it's worth checking their code to see how they achieved it. That could also be a good way to check if the descriptors are setup correctly.

Here's the code: https://github.com/midibox/mios32/blob/master/mios32/STM32F4xx/mios32_usb_midi.c
(They also have USB host support, btw)

stephenhensley:
@TheSlowGrowth this does not implement multiple virtual ports (yet). 

Does seem like it wouldn't be too crazy to go the route that they did on the mios32 stuff you linked.

This version 'just works'™️  for now, though it's still mostly wrapped into the CDC class. 

I'm pretty partial to cleaning this up and getting it in there for now with the known issue that you can only use it for MIDI _or_ CDC.

Sooner than later, I'd like to revisit the existing USB implementation, clean it up quite a bit, and starting working toward a (reasonably easily configurable) composite device class that can do both (and more) at once.

bolau:
Multiple virtual ports would be really nice, I agree. The Teensy also has it in the MIDI-over-USB implementation.

CorvusPrudens:
Actually, now that I think of it, adding multiple virtual ports should be pretty trivial. All it requires is some additions to the USB MIDI descriptor and an argument in the USB MIDI send / receive methods to select the port.

PR #367, opened 2021-06-30
CPP-ify sdram

beserge:
Created SdramHandler class.
Updated all of those typedef structs to enum classes.
Renamed and slightly tweaked functions to work within class.
Updated seed board support to use cpp-ified sdram.

### DIFF
--- a/src/daisy_seed.cpp
+++ b/src/daisy_seed.cpp
@@ -110,7 +110,6 @@ void DaisySeed::Init(bool boost)
     boost ? syscfg.Boost() : syscfg.Defaults();
     system.Init(syscfg);
 
-    dsy_sdram_init(&sdram_handle);
     dsy_qspi_init(&qspi_handle);
     dsy_gpio_init(&led);
     dsy_gpio_init(&testpoint);
@@ -207,10 +206,14 @@ void DaisySeed::SetTestPoint(bool state)
 
 void DaisySeed::ConfigureSdram()
 {
-    dsy_gpio_pin *pin_group;
-    sdram_handle.state             = DSY_SDRAM_STATE_ENABLE;
-    pin_group                      = sdram_handle.pin_config;
-    pin_group[DSY_SDRAM_PIN_SDNWE] = dsy_pin(DSY_GPIOH, 5);
+    dsy_gpio_pin *      pin_group;
+    SdramHandle::Config conf;
+
+    conf.state                              = SdramHandle::State::ENABLE;
+    pin_group                               = conf.pin_config;
+    pin_group[(int)SdramHandle::Pin::SDNWE] = dsy_pin(DSY_GPIOH, 5);
+
+    sdram_handle.Init(conf);
 }
 void DaisySeed::ConfigureQspi()
 {
--- a/src/daisy_seed.cpp
+++ b/src/daisy_seed.cpp
@@ -208,8 +208,7 @@ void DaisySeed::SetTestPoint(bool state)
 void DaisySeed::ConfigureSdram()
 {
     sdram_config.state = SdramHandle::State::ENABLE;
-    sdram_config.pin_config[(int)SdramHandle::Pin::SDNWE]
-        = dsy_pin(DSY_GPIOH, 5);
+    sdram_config.sdnwe = dsy_pin(DSY_GPIOH, 5);
 }
 void DaisySeed::ConfigureQspi()
 {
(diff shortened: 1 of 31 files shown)


beserge:
[SDRAM C++ wrapper](https://trello.com/c/Ie6qx1Kq/136-sdram-c-wrapper)

beserge:
Simply CPP-ifies SDRAM. There will be some more changes made on the GPIO branch (specifically changing the GPIO usage to the cpp gpio). Saved that for that pr, since it's outside the scope of this one.  Tested working on the patch multidelay example.

src/daisy_seed.cpp
```diff
@@ -207,10 +207,9 @@ void DaisySeed::SetTestPoint(bool state)
 
 void DaisySeed::ConfigureSdram()
 {
-    dsy_gpio_pin *pin_group;
-    sdram_handle.state             = DSY_SDRAM_STATE_ENABLE;
-    pin_group                      = sdram_handle.pin_config;
-    pin_group[DSY_SDRAM_PIN_SDNWE] = dsy_pin(DSY_GPIOH, 5);
+    sdram_config.state = SdramHandle::State::ENABLE;
+    sdram_config.pin_config[(int)SdramHandle::Pin::SDNWE]
```
TheSlowGrowth, line 211: For stronger type safety I would suggest to have each pin as a separate member in the pin_config struct, so that you can leave out the array indexing like this:
```cpp
sdram_config.pin_config.sdnwe = dsy_pin(DSY_GPIOH, 5);
```

TheSlowGrowth (changes requested):

stephenhensley:
Realistically, I think we can scrap the single-configured pin and hard code it. There was a difference on one old version of hardware board. 

Debatably, this doesn't need a config struct at all. I have not personally "disabled" the SDRAM, and I can't think of a reason that you'd need to do that at the config level. 

Plus, in the near future, I'd like to move the SDRAM initialization into the startup file. So that the text section can be used in some capacity, or for alternate linkers where the SDRAM is the primary memory for things where tons of memory is needed, but performance doesn't matter.

src/dev/sdram.h
```diff
@@ -55,19 +55,19 @@ SECTIONS
...
-#ifdef __cplusplus
-extern "C"
-{
-#endif
 #include <stdint.h>
 #include "daisy_core.h"
+extern "C"
+{
+#include "util/hal_map.h"
+}
```
stephenhensley, line 63: It looks like these can move inside the implementation file.

afaik, the hal map files have not had to be included in any headers yet, and I'd like to keep it that way. It prevents any external reliance on the HAL itself unless the user explicitly includes it.

src/dev/sdram.cpp
```diff
@@ -37,33 +36,24 @@ typedef struct
...
     {
-        if(sdram_periph_init() != DSY_SDRAM_OK)
-        {
-            return DSY_SDRAM_ERR;
-        }
-        if(sdram_device_init() != DSY_SDRAM_OK)
-        {
-            return DSY_SDRAM_ERR;
-        }
+        return Result::OK;
```
stephenhensley, line 51: Looks like this should return `ERR` instead.

stephenhensley (changes requested):
Just some little things.

Also, should update the CMakeLists.txt to have `sdram.cpp` instead of `sdram.c`

CorvusPrudens:
It looks like there's a small bug with the timing configuration that's stuck around since the base C version. In `sdram.cpp`, line 80:
```c++
SdramTiming.RPDelay              = 0;
```
the row precharge delay parameter is set to 0. If you dig through the HAL, you'll find that the function that actually writes this value to the FMC SDTR register subtracts one before shifting and ORing:
```c++
MODIFY_REG(Device->SDTR[FMC_SDRAM_BANK1],
               SDTR_CLEAR_MASK,
               (((Timing->LoadToActiveDelay) - 1U)                                      |
                (((Timing->ExitSelfRefreshDelay) - 1U) << FMC_SDTRx_TXSR_Pos) |
                (((Timing->SelfRefreshTime) - 1U)      << FMC_SDTRx_TRAS_Pos) |
                (((Timing->RowCycleDelay) - 1U)        << FMC_SDTRx_TRC_Pos)  |
                (((Timing->WriteRecoveryTime) - 1U)    << FMC_SDTRx_TWR_Pos)  |
                (((Timing->RPDelay) - 1U)              << FMC_SDTRx_TRP_Pos)  |
                (((Timing->RCDDelay) - 1U)             << FMC_SDTRx_TRCD_Pos)));
```
meaning that it actually underflows and writes 0xF to RPDelay and RCDDelay. I'm guessing it would only make a small different to performance, but we'll probably want to correct that. The HAL function is expecting the number of cycles of delay, not the actual register value, so a 1 should be written instead.

stephenhensley:
closing because I think most/all of the changes were absorbed into #390

PR #378, opened 2021-07-10
Finish MidiHandler::StartReceive comment

SSteve:
The comment for MidiHandler::StartReceive is incomplete. I'm assuming it should say the queue can be checked with HasEvents().


stephenhensley:
Thanks! Good catch!

PR #384, opened 2021-07-19
Cmake: Reslove LTO linker issue

brbrr:
When building my project with Cmake in `Release` mode I get the following error:

```
[build] [243/245  99% :: 7.613] Linking CXX static library libdaisy/libdaisy.a
[build] /usr/local/bin/arm-none-eabi-ar: libdaisy/CMakeFiles/daisy.dir/src/sys/dma.c.obj: plugin needed to handle lto object
[build] /usr/local/bin/arm-none-eabi-ranlib: libdaisy/libdaisy.a(dma.c.obj): plugin needed to handle lto object
```

With this small change, I can successfully build my project.

### DIFF
--- a/cmake/toolchains/stm32h750xx.cmake
+++ b/cmake/toolchains/stm32h750xx.cmake
@@ -28,10 +28,17 @@ endif()
 
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
 set(MCU "-mcpu=cortex-m7 -mfpu=fpv5-d16 -mfloat-abi=hard -mthumb")
-set(OBJECT_GEN_FLAGS "${MCU} -fno-builtin -fno-exceptions -Wall -Werror -ffunction-sections -fdata-sections -fomit-frame-pointer -finline-functions -Wno-attributes -Wno-strict-aliasing -Wno-maybe-uninitialized -Wno-missing-attributes -Wno-stringop-overflow -ffat-lto-objects")
+set(OBJECT_GEN_FLAGS "${MCU} -fno-builtin -fno-exceptions -Wall -Werror -ffunction-sections -fdata-sections -fomit-frame-pointer -finline-functions -Wno-attributes -Wno-strict-aliasing -Wno-maybe-uninitialized -Wno-missing-attributes -Wno-stringop-overflow")
 set(CMAKE_C_FLAGS   "${OBJECT_GEN_FLAGS} -std=gnu99 " CACHE INTERNAL "C Compiler options")
 set(CMAKE_CXX_FLAGS "${OBJECT_GEN_FLAGS} -Wno-register" CACHE INTERNAL "C++ Compiler options")
 set(CMAKE_ASM_FLAGS "${OBJECT_GEN_FLAGS} -x assembler-with-cpp " CACHE INTERNAL "ASM Compiler options")
+
+# Ensure the ar plugin is loaded (needed for LTO)
+set(CMAKE_AR ${TOOLCHAIN_BIN_DIR}/${TOOLCHAIN}-gcc-ar)
+set(CMAKE_C_ARCHIVE_CREATE "<CMAKE_AR> qcs <TARGET> <LINK_FLAGS> <OBJECTS>")
+set(CMAKE_C_ARCHIVE_FINISH   true)
+set(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> qcs <TARGET> <LINK_FLAGS> <OBJECTS>")
+set(CMAKE_CXX_ARCHIVE_FINISH   true)
 
 add_compile_definitions(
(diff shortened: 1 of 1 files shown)


TheSlowGrowth:
trying to understand the implications of this change...
According to [this question](https://stackoverflow.com/questions/13799452/what-is-the-difference-in-gcc-between-lto-and-fat-lto-objects) we should see worse compilation performance. Have you tried it with the slim object setting?

brbrr:
yeah, to be honest, I wasn't sure what I was doing, but apparently `-ffat-lto-objects` did resolve the build issue. 

I tried with `-fno-fat-lto-objects` flag, but no luck.

I forgot to mention that the build actually succeeds (exit code 0) but produces 2kb .bin file, which obviously not usable:

```
[build] [243/245  99% :: 8.118] Linking CXX static library libdaisy/libdaisy.a
[build] /usr/local/bin/arm-none-eabi-ar: libdaisy/CMakeFiles/daisy.dir/src/sys/dma.c.obj: plugin needed to handle lto object
[build] /usr/local/bin/arm-none-eabi-ranlib: libdaisy/libdaisy.a(dma.c.obj): plugin needed to handle lto object
[build] [244/245  99% :: 8.166] Linking CXX static library libWoopyOneLib.a
[build] /usr/local/bin/arm-none-eabi-ar: CMakeFiles/WoopyOneLib.dir/src/init.cpp.obj: plugin needed to handle lto object
[build] /usr/local/bin/arm-none-eabi-ranlib: libWoopyOneLib.a(init.cpp.obj): plugin needed to handle lto object
[build] [245/245 100% :: 8.704] Linking CXX executable WoopyOne.elf
[build] /usr/local/Caskroom/gcc-arm-embedded/10-2020-q4-major/gcc-arm-none-eabi-10-2020-q4-major/bin/../lib/gcc/arm-none-eabi/10.2.1/../../../../arm-none-eabi/bin/ld: warning: cannot find entry symbol Reset_Handler; defaulting to 0000000008000000
[build] Build finished with exit code 0
```

recursinging:
I think this might be related to a similar problem I was having with the templated OLED classes - albeit with a different linker error.  I was able to reproduce it by simply including an OLED instance in a daisyHat test and building with `Release`.  I never got around to removing flags to localize the problem.  I'll try and verify later today.

recursinging:
OK, I think I got it figured out...

Indeed this is the same issue I ran into, and this PR is indeed one way to solve the problem. 

So CMake in `Release` mode enables the `-flto` compiler flag, which according to [this doc](https://gcc.gnu.org/onlinedocs/gccint/LTO-Overview.html#LTO-Overview):

> By default, object files generated with LTO support contain only GIMPLE bytecode. Such objects are called “slim”, and they require that tools like ar and nm understand symbol tables of LTO sections. For most targets these tools have been extended to use the plugin infrastructure, so GCC can support “slim” objects consisting of the intermediate code alone.

...as you can see in @brbrr 's error message: `plugin needed to handle lto object`

So apparently gcc is not being correctly instructed on where to find the required plugin(s). Indeed if we continue reading the docs:

> GIMPLE bytecode could also be saved alongside final object code if the -ffat-lto-objects option is passed, or if no plugin support is detected for ar and nm when GCC is configured.

...this explains why adding`-ffat-lto-objects` solves this problem... but what if we don't want fat objects? I mean slim is always better than fat right? ;-)

Well, I found a solution for configuring the plugin with CMake [here](https://stackoverflow.com/questions/39236917/using-gccs-link-time-optimization-with-static-linked-libraries), and tada:

```
# Ensure the ar plugin is loaded (needed for LTO)
set(CMAKE_AR ${TOOLCHAIN_BIN_DIR}/${TOOLCHAIN}-gcc-ar)
set(CMAKE_C_ARCHIVE_CREATE "<CMAKE_AR> qcs <TARGET> <LINK_FLAGS> <OBJECTS>")
set(CMAKE_C_ARCHIVE_FINISH   true)
set(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> qcs <TARGET> <LINK_FLAGS> <OBJECTS>")
set(CMAKE_CXX_ARCHIVE_FINISH   true)
```

...this fixes the problem without needing the `-ffat-lto-objects` flag.

brbrr:
whoa, that's great research!

It did work in my project, so I pushed your snipped into this PR.

stephenhensley:
Looks good! Thanks for the contribution, and diligent research everyone!

I still haven't had a chance to start working on any real projects with the new CMake build, but I'm excited to start using it on the next project I start!

@brbrr we've started to add some changes to the CHANGELOG.md file as we go. So I'd say go ahead an add a line in there for this change before we sign off and merge it, but we already have this line from another fix the other day:

>build: small fixes in CMakeLists.txt

So I think that generically covers this as well, but something to be aware of for any future PRs :)

PR #390, opened 2021-08-11
Bootloader additions

CorvusPrudens:
This PR includes SDRAM startup, as well as QSPI and bootloader changes. The bootloader procedure is documented (and should be made available quite soon).


stephenhensley:
My latest few commits temporarily removed the startup code for Sdram Init, and added it back in so that a "non-volatile" program (i.e. located in internal flash) would still initialize the SDRAM within the `DaisySeed::Init`.

Testing this with apps running directly from the internal flash, and using the new bootloader to run in AXI SRAM seems to be solid.

With the SDRAM Init only happening in the startup (before the clock tree gets configured), there were a few issues. The SDRAM would start out working, and begin to gradually fail (on reverb this was audible as glitch-like artifacts and then either a failure to read data from the ram or all of the audio would stop). 

So if we want to integrate the SDRAM init into the startup I think we'll probably want to set up the clock tree there as well. That'll save any potential issues with reconfiguring the clocks afterwards.

For now, though the bootloader stuff is working with these few changes (including using the SDRAM memory for the heap, for things like pd2dsy).

I haven't gone through and done a full review yet either, but at least everything's working.

CorvusPrudens:
# Summary of Changes

## PRs incorporated
- sdram cpp
  - This requires changes to the new patch_sm code
- qspi cpp

## File Changes
- CMakeLists.txt -- adjusted for sdram cpp
- stm32h7xx_hal.c -- commented out `HAL_DeInit` code that causes system restart
- root Makefile -- added `program-app` target and made `LDSCRIPT` optionally assignable
- sdram_startup.h -- added startup code that may not ever be used
- startup_stm32h750xx.c -- added `ifdef` that skips `SystemInit` function for bootloaded applications
- daisy_patch_sm -- updated sdram to cpp
- daisy_seed -- updated sdram to cpp, added runtime execution location checks, and added deinit method
- codec_ak4556,  audio,  usb, sai, tim, dma, fatfs, system -- added deinit methods / functions
- system -- added `GetProgramMemory` method to check current address of code execution, with accompanying `constexpr`s, as well as a config variable for optional clock configuration (which is unnecessary for bootloaded applications)
- usbd_conf.h -- set `USBD_SUPPORT_USER_STRING` macro to 1 so the DFU code can specify QSPI's address and size

Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal.c
```diff
@@ -155,8 +155,8 @@ HAL_StatusTypeDef HAL_Init(void)
 HAL_StatusTypeDef HAL_DeInit(void)
 {
   /* Reset of all peripherals */
-  __HAL_RCC_AHB3_FORCE_RESET();
-  __HAL_RCC_AHB3_RELEASE_RESET();
+  // __HAL_RCC_AHB3_FORCE_RESET();
+  // __HAL_RCC_AHB3_RELEASE_RESET();
```
stephenhensley, line 159: Are you using something in AHB3 during the bootloader stuff that you can't reset? Or is this just hanging around from when you were trying to get the jump to work in the first place?

core/sdram_startup.h
```diff
@@ -0,0 +1,519 @@
+#include <stdint.h>
```
stephenhensley, line 1: Is this file used anymore? Or just vestigial from when we wanted to move SDRAM to startup.

core/startup_stm32h750xx.c
```diff
@@ -5,6 +5,7 @@
 */
 
 #include <stddef.h>
+//#include "sdram_startup.h"
```
stephenhensley, line 8: 🧹

core/startup_stm32h750xx.c
```diff
@@ -1547,8 +1548,13 @@ void __attribute__((naked, noreturn)) Reset_Handler()
 	for (pDest = &_sbss; pDest != &_ebss; pDest++)
 		*pDest = 0;
 
+	#ifndef BOOT_APP
 	SystemInit();
+	//SdramInit();
```
stephenhensley, line 1553: 🧹

src/daisy_patch_sm.cpp
```diff
@@ -227,9 +227,10 @@ namespace patch_sm
         system.Init(syscfg);
         /** Memories */
         /** FMC SDRAM */
-        sdram.state                           = DSY_SDRAM_STATE_ENABLE;
-        sdram.pin_config[DSY_SDRAM_PIN_SDNWE] = {DSY_GPIOH, 5};
-        dsy_sdram_init(&sdram);
+        // sdram.state                           = DSY_SDRAM_STATE_ENABLE;
+        // sdram.pin_config[DSY_SDRAM_PIN_SDNWE] = {DSY_GPIOH, 5};
+        // dsy_sdram_init(&sdram);
```
stephenhensley, line 232: 🧹

src/daisy_seed.cpp
```diff
@@ -1,4 +1,5 @@
 #include "daisy_seed.h"
+#include "stm32h7xx_hal.h"
```
stephenhensley, line 2: Do you still need this for something?

If not, lets get rid of that, otherwise let's update it to the hal_map file that way there's a single include for any libDaisy sources that need access to the HAL (I still dream of a HAL-less future).

src/dev/sdram.cpp
```diff
@@ -1,7 +1,9 @@
 #include <stm32h7xx_hal.h>
```
stephenhensley, line 1: I think the HAL file is included through the hal_map file.

src/dev/sdram.cpp
```diff
@@ -304,54 +333,47 @@ static void HAL_FMC_MspInit(void)
...
+    //        GPIO_InitStruct.Pin = GPIO_PIN_5;
+    //        GPIO_InitStruct.Mode = GPIO_MODE_AF_PP;
+    //        GPIO_InitStruct.Pull = GPIO_NOPULL;
+    //        GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_VERY_HIGH;
+    //        GPIO_InitStruct.Alternate = GPIO_AF12_FMC;
+    //        HAL_GPIO_Init(GPIOH, &GPIO_InitStruct);
+    //        break;
+    //    default:
+    //        break;
+    //    }
```
stephenhensley, line 376: We can get rid of this at this point -- we don't handle different boards this way anymore, and aside from some abandoned/out of date hardware all boards use the same RAM pins now.

src/dev/sdram.cpp
```diff
@@ -492,15 +514,15 @@ void HAL_SDRAM_MspDeInit(SDRAM_HandleTypeDef *sdramHandle)
...
+    //    for (pSource = &_sisdram_data, pDest = &_ssdram_data; pDest != &_esdram_data; pSource++, pDest++)
+    //        *pDest = *pSource;
     //
-    //	for (pDest = &_ssdram_bss; pDest != &_esdram_bss; pDest++)
-    //		*pDest = 0;
+    //    for (pDest = &_ssdram_bss; pDest != &_esdram_bss; pDest++)
+    //        *pDest = 0;
 }
 
 //Program the <project name>_QSPIFLASH.bin file to QSPIFLASH manually
```
stephenhensley, line 528: Is this used/needed? if not lets clean it up.

src/hid/audio.cpp
```diff
@@ -32,6 +32,7 @@ class AudioHandle::Impl
     AudioHandle::Result Init(const AudioHandle::Config config, SaiHandle sai);
     AudioHandle::Result
                         Init(const AudioHandle::Config config, SaiHandle sai1, SaiHandle sai2);
+    AudioHandle::Result Deinit();
```
stephenhensley, line 35: I don't want to mark every single one, but I think `DeInit` is a bit more common for naming convention than `Deinit`

src/per/pin.h
```diff
@@ -0,0 +1,47 @@
+#pragma once
```
stephenhensley, line 1: I don't see anything that uses this, and it seems like this some how got tied in or checked out from #354 

If it's not in use let's remove it until we work on that PR.

src/sys/system.h
```diff
@@ -152,10 +162,23 @@ class System
...
+
+    static constexpr uint32_t sram_start = 0x24000000U;
+    static constexpr uint32_t sram_end   = sram_start + 0x80000U;
+    static constexpr uint32_t qspi_start = 0x90040000U;
+    // TODO -- this is a bit too large:
+    static constexpr uint32_t qspi_end       = qspi_start + 0x800000U;
+    static constexpr uint32_t internal_start = 0x08000000U;
+    static constexpr uint32_t internal_end   = internal_start + 0x20000U;
+
+    static constexpr uint32_t expected_stack = 0x20020000U;
```
stephenhensley, line 181: constant naming conventions should be:
```c++
static constexpr uint32_t kSramStart = 0x24000000U;
```
etc. to make clear that it's constant while reading.

stephenhensley (changes requested):

Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal.c
CorvusPrudens: Thought I put a comment there, but I guess not. There wasn't any startup configuration / deinitialization routine I could put together that didn't result in the H750 just restarting immediately upon reaching those two lines of code. I'm not sure what underlying system interaction caused it, but it was very consistent. Since it defeats the purpose of deinitialization (why would you ever do it unless you wanted to avoid a reset?), I though it would be fine to comment it out entirely.

core/sdram_startup.h
CorvusPrudens: Yeah it can be removed.

src/dev/sdram.cpp
CorvusPrudens: As far as I can tell, it's not used at all.

CorvusPrudens:
[Cleanup PR for Bootloader Additions](https://trello.com/c/Qmw7pctl/248-cleanup-pr-for-bootloader-additions)

Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal.c
stephenhensley: Interesting. Yeah, could be something you're using during the bootloader that trips out on AHB3 reset. Since these aren't _our_ files, and we may upgrade, at some point. Can you make a note in the `Drivers/notes.md` file with the change. So that we don't forget to replicate it when we update HAL versions or anything.

src/sys/system.h
```diff
@@ -152,25 +162,29 @@ class System
...
+
+    static constexpr uint32_t kSramStart = 0x24000000U;
+    static constexpr uint32_t kSramEnd   = kSramStart + 0x80000U;
+    static constexpr uint32_t kQspiStart = 0x90040000U;
+
+    static constexpr uint32_t kQspiEnd       = kQspiStart + 0x7c0000;
+    static constexpr uint32_t kInternalStart = 0x08000000U;
+    static constexpr uint32_t kInternalEnd   = kInternalStart + 0x20000U;
+
+    static constexpr uint32_t kExpectedStack = 0x20020000U;
```
stephenhensley, line 181: Is this const used for anything? 

idk if we should expect the stack to be in one place or the other because A: we just made the default linker conditional. So users can move it elsewhere. and B: if we wanted access to this we could probably provide/wrap access around a linker symbol instead.

stephenhensley (changes requested):
The last few things I made notes about as well as:

* Adding memory-aware system init should be added to the DaisyPatchSM

src/sys/system.h
CorvusPrudens: Maybe I'm just bad, but I didn't find the linker variable approach very straightforward. Even after declaring the correct variables as `extern` within libdaisy, it seemed like the final compilation failed to actually link them correctly.

In any case, that approach wouldn't really work here because this is used by the bootloader to check _other_ binaries. Assuming the stack is located in the same place, checking the first four bytes is a convenient and reliable way to determine if a binary is likely correctly formatted and executable. I suppose that checking could be skipped, though. As long as the stack and entry point are within valid memory, I suppose it's fine.

src/sys/system.h
```diff
@@ -148,29 +162,41 @@ class System
...
+    /** Returns an enum representing the the memory region 
+     *  that the given address belongs to.
+     *  \param address The address to be checked
+     */
+    static MemoryRegion GetMemoryRegion(uint32_t address);
+
+    /** This offset from the base QSPI address is reserved
+     *  for libDaisy functionality.
+     */
+    static constexpr uint32_t kQspiOffset = 0x40000U;
```
stephenhensley, line 183: This almost makes it seem like users _shouldn't_ use the first 256kB, but it's actually that the first 0x40000 are left for user data.

src/sys/system.h
```diff
@@ -148,29 +162,41 @@ class System
...
+
+    // static constexpr uint32_t kSramStart = 0x24000000U;
+    // static constexpr uint32_t kSramEnd   = kSramStart + 0x80000U;
+    // static constexpr uint32_t kQspiStart = 0x90040000U;
+
+    // static constexpr uint32_t kQspiEnd       = kQspiStart + 0x7c0000;
+    // static constexpr uint32_t kInternalStart = 0x08000000U;
+    // static constexpr uint32_t kInternalEnd   = kInternalStart + 0x20000U;
+
+    // static constexpr uint32_t kExpectedStack = 0x20020000U;
```
stephenhensley, line 193: 🧹

CHANGELOG.md
```diff
@@ -5,20 +5,27 @@
...
 * qspi: updated from C to C++, and fixed up the API to be cleaner, and simpler to use
+* sdram: Create SdramHandle class. Gets rid of configurable pin and hardcodes it instead.
 
 ### Features
 
 * adc: added initialization mappings for pins ADC1_INP12 (PC2) and ADC1_INP13 (PC3) (Not accessible on Daisy Seed)
+* bootloader: BOOT_APP macro added for running a program from the daisy bootloader
+* bootloader: added a number of deinitialization methods to core classes
+* bootloader: added `System` and `DaisySeed` code to check code location at runtime and initialize accordingly
+* bootloader: added `program-app` make target to upload code to the daisy bootloader
```
stephenhensley, line 16: I'd classify "features" as things that are user-accessible. so you could probably move these to the "other" section instead.

might be worth lumping most of these together into something like:

* bootloader: added deinit, and other groundwork for compatibility with Daisy Bootloader

We'll be making a wiki page dedicated to all things bootloader as we wrap up the project. So I don't think we need to bloat the changelog too much.

CHANGELOG.md
```diff
@@ -5,20 +5,27 @@
...
 * spi: added DMA Transactions (same type of queue system as I2C) to the SPI Handle class.
+* sdram: Update from c to cpp.
 * util: added new PersistentStorage class for storing/recalling data between power cycles
 * util: added new VoctCalibration helper class for calibrating ADC inputs
 * seed: added support for Daisy Seed 1.1 (aka Daisy Seed rev5) hardware. Pin-compatible, with same form factor. WM8731 Codec instead of AK4556.
 
 ### Bug fixes
 
 * adc: fixed bug with ADC where designs with more than 8 channels AND a mux would halt the ADC.
+* Change SdramTiming.RPDelay to 1 to avoid underflow bug.
```
stephenhensley, line 28: add: `sdram:` to beginning of this item.

stephenhensley (changes requested):
Final changes, most of which are just cleaning up the change log, and removing old comments.

stephenhensley (approved):

src/sys/system.h
stephenhensley: you're not bad, haha. There's probably something weird about them that makes it trickier to deal with.  I think I had something reading one of them at some point a long time ago when I was doing some QSPI prior to the daisy release, but I can't remember if/what I had to do to actually read them.

PR #394, opened 2021-09-04
Fix for occasional qspi init bug

CorvusPrudens:
HAL strikes again. The `NbData` member of `QSPI_CommandTypeDef` is subtracted by 1 before being used. This either causes the initialization to take longer than it should when NbData is set to 0, or can cause a hard fault if the HAL code tries to read invalid data (like if the data buffer is on the stack, which is usually pretty close to the end of DTCMRAM). 

I don't think the `EnterQuadMode`, `ExitQuadMode`, and `EraseSector` ones are really that important (since the HAL ignores them anyway), but it's probably best to set them to 1.

### DIFF
--- a/src/per/qspi.cpp
+++ b/src/per/qspi.cpp
@@ -360,7 +360,7 @@ QSPIHandle::Result QSPIHandle::Impl::EraseSector(uint32_t address)
     s_command.AlternateByteMode = QSPI_ALTERNATE_BYTES_NONE;
     s_command.DataMode          = QSPI_DATA_NONE;
     s_command.DummyCycles       = 0;
-    s_command.NbData            = 0;
+    s_command.NbData            = 1;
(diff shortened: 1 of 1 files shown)


CorvusPrudens:
[C++-ify QSPI peripheral code](https://trello.com/c/hmQiSNzI/123-c-ify-qspi-peripheral-code)

stephenhensley:
Interesting catch.

Weirdly, only the QSPI mode enable one (where DataMode is != `QSPI_DATA_NONE`) would be the only one that could trigger the issue. (The HAL Commands seem to ignore the Command->NbData unless the DataMode is something other than none). That said, 0 shouldn't be used anyway. 

Perhaps more interesting, `0xffffffff` as the `NbData-1` would actually be set to the specified size of the flash chip. So probably best to update them all to 1 like you did.

PR #396, opened 2021-09-07
Feature - Persistent Storage

stephenhensley:
This is a starting point for an abstraction around storing/recalling data between power cycles.

We'll be adding another set of helpers shortly that will be calibration-specific, but this is the core class needed for storing/recalling stuff.

Usage involves creating a struct to hold whatever settings you want, using that as the "SettingsStruct" for this class. 

There are a few todo items that would make this better, but it's functional, and useful as it is.

The "blocking" time isn't really too bad, but if you have constant LED patterns going, or are updating an OLED fast (like for an oscope), the stutter when it saves is noticeable.

Here's a brief (pseudocode) example:

```c++
#include "daisy_seed.h"

/** Custom struct for  data to save. 
  *  The defaults can be set in the constructor,
  *  the initialization list, or manually set before
  *  calling the "PesistentStorage::Init" function.
  */
struct MySettings : a(false), b(2.f), c(0.f) 
{ 
    bool a;
    float b, c;
};

/** Normal Daisy Seed global Hardware object */
DaisySeed hw;

/** New Persistent Memory Manager */
PersistentStorage<MySettings> setting_storage;

int main()
{
    hw.Configure();
    hw.Init();

    /** Settings are already initialized due to Constructor */
    MySettings settings;
    /** Initialize the storage, but saving the default values to the location
          or loading the data that's been saved previously into memory */
    setting_storage.Init(&hw.qspi, settings);

    /** At this point "settings" contains either the defaults, or the last-saved data
     *   So we can assign them to where ever they need to go within the application.
     */
   mysynth.SetA(settings.a);
   mysynth.SetB(settings.b);
   mysynth.SetC(settings.c);

    while(1) 
    {
        /** This hypothetical flag would get set by the application to 
             to signal that a save is ready, though you could just save any
             time there's a change */
        if (readyToSave) {
            /** Collect the data from where ever in the application it is */
            settings.a = mysynth.GetA();
            settings.b = mysynth.GetB();
            settings.c = mysynth.GetC();

            /** And trigger the save */
            setting_storage.Save();
        }
    }
}

```

### DIFF
--- a/src/per/qspi.h
+++ b/src/per/qspi.h
@@ -1,6 +1,8 @@
 #ifndef DSY_QSPI
 #define DSY_QSPI /**< Macro */
 
+#ifndef UNIT_TEST // for unit tests, a dummy implementation
+
 #include <cstdint>
 #include "daisy_core.h" // Added for dsy_gpio_pin typedef
 
@@ -166,4 +168,71 @@ class QSPIHandle
 
 } // namespace daisy
 
-#endif
+#else
+
+#include <cstdint>
+#include "../tests/TestIsolator.h"
+
+namespace daisy
+{
+/** This is a dummy implementation for use in tests. 
+ *  In your tests you can use this as a placeholder 
+ *  for the physical volatile memory. 
+ *  This provides a block of memory that can be erased, or written
+ *  to.
+ *  
+ *  On the hardware, the memory can be read directly from addresses
+ *  0x90000000 and above. However, within the unit tests you must use
+ *  the Read() function.
+ */
+class QSPIHandle
+{
+  public:
+    enum Result
+    {
+        OK = 0,
+        ERR
+    };
+
+    static Result Write(uint32_t address, uint32_t size, uint8_t* buffer)
+    {
+        /** 256-byte aligned, normalized address value */
+        uint32_t adjusted_addr = (address - kBaseAddress) & (uint32_t)(~0xff);
+        std::copy(&buffer[adjusted_addr],
+                  &buffer[adjusted_addr + size],
+                  (uint8_t*)testIsolator_.GetStateForCurrentTest()->memory_);
+        return Result::OK;
+    }
+
+    static Result Erase(uint32_t start_addr, uint32_t end_addr)
+    {
+        uint32_t adjusted_start_addr
+            = (start_addr - kBaseAddress) & (uint32_t)(~0xff);
+        uint32_t adjusted_end_addr
+            = (end_addr - kBaseAddress) & (uint32_t)(~0xff);
+        uint32_t* buff = testIsolator_.GetStateForCurrentTest()->memory_;
+        // Erases memory by setting all bits to 1
+        std::fill(
+            &buff[adjusted_start_addr], &buff[adjusted_end_addr], 0xffffffff);
+        return Result::OK;
+    }
+
+  private:
+    static constexpr uint32_t kBaseAddress = 0x90000000;
+    struct QSPIState
+    {
+        // 8MB pool of memory -- not sure if we should limit this,
+        // but it will allow for testing out of range read/writes..
+        // Also, I think this should probably be in the heap instead.
+        uint32_t memory_[1024 * 1024 * 8 / 4] = {0xffffffff};
+    };
+    static TestIsolator<QSPIState> testIsolator_;
+};
+
+
+} // namespace daisy
+
+
+#endif // ifndef UNIT_TEST
+
+#endif // ifndef DSY_QSPI
--- a/src/per/qspi.h
+++ b/src/per/qspi.h
@@ -154,6 +154,12 @@ class QSPIHandle
      */
     Status GetStatus();
 
+    /** Returns a pointer to the actual memory used 
+     *  The memory at this address is read-only
+     *  to write to it use the Write function.
+    */
+    void* GetData();
+
     QSPIHandle() : pimpl_(nullptr) {}
     QSPIHandle(const QSPIHandle& other) = default;
     QSPIHandle& operator=(const QSPIHandle& other) = default;
@@ -180,10 +186,6 @@ namespace daisy
  *  for the physical volatile memory. 
  *  This provides a block of memory that can be erased, or written
  *  to.
- *  
- *  On the hardware, the memory can be read directly from addresses
- *  0x90000000 and above. However, within the unit tests you must use
- *  the Read() function.
  */
 class QSPIHandle
 {
@@ -208,7 +210,7 @@ class QSPIHandle
     static Result Write(uint32_t address, uint32_t size, uint8_t* buffer)
     {
         // 256-byte aligned, normalized address value
-        uint32_t adjusted_addr = (address - kBaseAddress) & (uint32_t)(~0xff);
+        uint32_t adjusted_addr = (address) & (uint32_t)(~0xff);
         // Make sure memory is of approriate size
         uint32_t total_bytes = adjusted_addr + size;
         AdaptToSize(total_bytes);
@@ -221,10 +223,8 @@ class QSPIHandle
 
     static Result Erase(uint32_t start_addr, uint32_t end_addr)
     {
-        uint32_t adjusted_start_addr
-            = (start_addr - kBaseAddress) & (uint32_t)(~0xff);
-        uint32_t adjusted_end_addr
-            = (end_addr - kBaseAddress) & (uint32_t)(~0xff);
+        uint32_t adjusted_start_addr = (start_addr) & (uint32_t)(~0xff);
+        uint32_t adjusted_end_addr   = (end_addr) & (uint32_t)(~0xff);
 
         // guard addresses
         assert(adjusted_start_addr < kMaxAdjustedAddr);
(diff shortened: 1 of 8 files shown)


stephenhensley:
Something I overlooked is that the way this is currently set up it will require that the user write custom `!=operator` for their struct.

I wrongly assumed that default comparison objects were made for structs, but I just realized this is not the case.

We could use something like memcmp to check them instead to remove that requirement, but I've enjoyed being able to be specific about what "inequality" would actually trigger a save, (i.e. using a mode-change to toggle saving all parameters rather than constantly saving anytime a parameter changes).

src/util/PersistentStorage.h
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /** Restores the settings stored in the QSPI */
+    void RestoreDefaults()
+    {
+        *settings_ = default_settings_;
+        state_     = State::FACTORY;
+        StoreSettings();
+    }
+
```
TheSlowGrowth, line 86:
```suggestion
    SettingStruct& GetSettings() { return settings_; }       
```

src/util/PersistentStorage.h
```diff
@@ -0,0 +1,119 @@
...
+        {
+            qspi_->Erase(start_address_, start_address_ + sizeof(s));
+            qspi_->Write(start_address_, sizeof(s), (uint8_t *)&s);
+        }
+    }
+
+    QSPIHandle *   qspi_;
+    uint32_t       start_address_;
+    SettingStruct  default_settings_;
+    SettingStruct *settings_;
```
TheSlowGrowth, line 114:
```suggestion
    SettingStruct settings_;
```

src/util/PersistentStorage.h
```diff
@@ -0,0 +1,119 @@
...
+     *      This defaults to the first address on the chip, and should be a multiple of 256
+     **/
+    void Init(QSPIHandle *   qspi,
+              SettingStruct &settings,
+              uint32_t       start_address = 0x90000000)
+    {
+        qspi_             = qspi;
+        state_            = State::UNKNOWN;
+        default_settings_ = settings; // copy data to fixed struct.
+        settings_ = &settings; // copy reference to pointer for modification.
```
TheSlowGrowth, line 50:
```suggestion
    void Init(QSPIHandle&          qspi,
              const SettingStruct& defaults,
              uint32_t             start_address = kMemBaseAddr)
    : qspi_(qspi)
    , start_address_(start_address)
    , default_settings_(defaults)
    , settings_(defaults)
    , state_(State::UNKNOWN)
    {
```

src/util/PersistentStorage.h
```diff
@@ -0,0 +1,119 @@
...
+        // Use the `==operator` in custom SettingStruct to fine tune
+        // what may or may not trigger the erase/save.
+        if(*settings_ != in_place_storage_->user_data)
+        {
+            qspi_->Erase(start_address_, start_address_ + sizeof(s));
+            qspi_->Write(start_address_, sizeof(s), (uint8_t *)&s);
+        }
+    }
+
+    QSPIHandle *   qspi_;
```
TheSlowGrowth, line 111:
```suggestion
    QSPIHandle&   qspi_;
```

src/util/PersistentStorage.h
```diff
@@ -0,0 +1,119 @@
...
+        if(cur_state != State::FACTORY && cur_state != State::USER)
+        {
+            // Initialize the Data store State::FACTORY, and the DefaultSettings
+            state_ = State::FACTORY;
+            StoreSettings();
+        }
+        else
+        {
+            state_     = cur_state;
+            *settings_ = in_place_storage_->user_data;
```
TheSlowGrowth, line 65:
```suggestion
            settings_ = in_place_storage_->user_data;
```

src/util/PersistentStorage.h
```diff
@@ -0,0 +1,119 @@
...
+  private:
+    static constexpr uint32_t kMemBaseAddr = 0x90000000;
+
+    struct SaveStruct
+    {
+        State         storage_state;
+        SettingStruct user_data;
+    };
+
+    void StoreSettings()
```
TheSlowGrowth, line 96:
```suggestion
    void StoreSettingsIfChanged()
```

src/util/PersistentStorage.h
```diff
@@ -0,0 +1,119 @@
...
+    {
+        SaveStruct s;
+        s.storage_state = state_;
+        s.user_data     = *settings_;
+        // Only actually save if the new data is different
+        // Use the `==operator` in custom SettingStruct to fine tune
+        // what may or may not trigger the erase/save.
+        if(*settings_ != in_place_storage_->user_data)
```
TheSlowGrowth, line 104:
```suggestion
        if((settings_ != in_place_storage_->user_data) || (state_ != in_place_storage_->storage_state))
```

TheSlowGrowth (changes requested):
Since the settings struct is a template argument, why not make the settings a member of this class? This makes it easier to use (no need to make sure the settings struct stays alive - no errors because people were passing in temporary variables). It also removes the double-meaning of the `settings` argument in the constructor. 

I added a couple of suggestions based around this idea.

TheSlowGrowth:
Some tests for this would be amazing :-)

stephenhensley:
@TheSlowGrowth thanks for the thorough review! I'll address the requested changes.

src/util/PersistentStorage.h
stephenhensley: woah, I didn't know you could initializer lists in non-constructor functions. that's interesting.

TheSlowGrowth: Oopsi. I was wrong, I misinterpreted this as the constructor. Doh. Sorry ...

src/util/PersistentStorage.h
stephenhensley: Hmm, this would auto-trigger a save the first time the save function gets called by an application, even if all values are still the defaults.

Not sure if that would be desired or not. Question is when/if an app checks "GetState" is it to see if it's still the factory-default values, or if a save has been performed at some point?

src/util/PersistentStorage.h
stephenhensley: Haha, no problem. I feel like I still learn about some new language feature every other week. So I wouldn't have been surprised if you could.

stephenhensley:
Addressed pretty much everything in the review so far. 

Still need to add tests, but there's an example for context in the comments of #397

src/per/qspi.h
```diff
@@ -166,4 +168,113 @@ class QSPIHandle
...
+{
+/** This is a dummy implementation for use in tests. 
+ *  In your tests you can use this as a placeholder 
+ *  for the physical volatile memory. 
+ *  This provides a block of memory that can be erased, or written
+ *  to.
+ *  
+ *  On the hardware, the memory can be read directly from addresses
+ *  0x90000000 and above. However, within the unit tests you must use
+ *  the Read() function.
```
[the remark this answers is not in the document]
TheSlowGrowth: I know this is not actually part of this PR, but maybe it's worth considering for a followup PR

src/util/PersistentStorage.h
```diff
@@ -0,0 +1,155 @@
...
+      in_place_storage_(nullptr)
+    {
+    }
+
+    /** Initialize Storage class
+     *
+     *  The values in this class will be stored as the default
+     *  for restoration to 'factory' settings.
+     *
+     *  \param qspi handle to the hardware qspi peripheral.
```
TheSlowGrowth, line 67: the param descriptions are not yet up to date

src/per/qspi.h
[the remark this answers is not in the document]
TheSlowGrowth: If we want to enable the `reinterpret_cast<MyType*>(start_address_)` logic (which can be quite nice!) we could also do it differently: Instead of providing a `Read` function, we could provide a `GetData()` function that returns a pointer to the base memory address.

tests/PersistentStorage_gtest.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+TEST(util_PersistentStorage, a_stateAfterInitClean)
+{
+    QSPIHandle       qspi;
+    StorageTestClass storage(qspi);
+    StorageTestData  defaults;
+
+    // Resets the emulated memory pool
+    qspi.ResetAndClear();
+
```
TheSlowGrowth, line 26:
```suggestion
```

tests/PersistentStorage_gtest.cpp
```diff
@@ -0,0 +1,87 @@
...
+TEST(util_PersistentStorage, a_stateAfterWrite)
+{
+    QSPIHandle       qspi;
+    StorageTestClass storage(qspi);
+    StorageTestData  defaults;
+
+    // Resets the emulated memory pool
+    qspi.ResetAndClear();
+
```
TheSlowGrowth, line 45:
```suggestion
```

tests/PersistentStorage_gtest.cpp
```diff
@@ -0,0 +1,87 @@
...
+TEST(util_PersistentStorage, c_recallData)
+{
+    QSPIHandle       qspi;
+    StorageTestClass storage(qspi);
+    StorageTestData  defaults;
+
+    // Resets the emulated memory pool
+    qspi.ResetAndClear();
+
```
TheSlowGrowth, line 69:
```suggestion
```

tests/PersistentStorage_gtest.cpp
```diff
@@ -0,0 +1,87 @@
...
+    // Initialize defaults (sets to FACTORY)
+    storage.Init(defaults);
+
+    //Write a new value and save (sets to USER)
+    auto &data = storage.GetSettings();
+    data.a     = 0;
+    storage.Save();
+
+    // Using a new instance to prevent any volatile bits from sticking through.
```
TheSlowGrowth, line 78: nice, didn't think of this, but it's a great way of testing this!

src/per/qspi.h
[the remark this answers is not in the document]
stephenhensley: The idea of a `Read()` function is nice, but I prefer the `GetData()` idea instead. That simplifies the actions needed here a bit.

That said, once we get through a few more of these PRs I want to go in and add non-blocking write functions that use the DMA. At that point, I could certainly see the benefits of adding Read function. It could be useful for bigger chunks of data (thinking wavetables, etc.)  to somewhere with faster access times prior to usage, as well as for profiling the memory itself. Which is a little more opaque when memory mapped.

I'll give adding the GetData function a go, and see how it looks.

tests/PersistentStorage_gtest.cpp
stephenhensley: If only it worked correctly! 😆 This is the one I still need to troubleshoot.

stephenhensley:
Well, I fixed some issues related to the offset with the mock-QSPI, while writing some tests to make sure it isn't causing this fail.

Re. the recall test thats failing. Craziest thing is that on my mac here, I have run the test that's failing probably 20-30 times and it's failed this way once. Every other time it passes.. So I'll troubleshoot some more when I can reproduce the fail a bit more.

PR #400, opened 2021-09-13
PatchSM - Simplifying user experience

stephenhensley:
All of these changes overall address the user experience of working with libDaisy, especially when being used by people new to C++.

* Added shorthand, `OUT_L`, `OUT_R`, `IN_L`, `IN_R` macros that can be used within the (non-interleaved) audio callback
* Added internal DAC callback to PatchSM board, and have generic setters for sending voltage instead of 12-bit value.
  * You can still override the callback for doing _real_ audio rate stuff, otherwise the default callback just copies the converted values to the outputs.
  * Adding some interpolation over the DAC blocksize would be nice. That way it smoothly blends from sample to sample over blocksize samples.
* Added a bunch of const members to  the class for the Pin mapping (i.e. `DaisyPatchSM::B4`, `DaisyPatchSM::C10`, etc.)
  * Once the C++ GPIO PR (#354) gets merged I think we can make these constexpr instead, but with the current pin-types this wasn't possible without modification
* Added members for the Gate Input and Gate Outputs on the Patch SM and pre-initialized them.

### DIFF
--- a/src/daisy_core.h
+++ b/src/daisy_core.h
@@ -47,19 +47,19 @@ cache enabled.
 
 /** shorthand macro for simplifying the reading of the left 
  *  channel of a non-interleaved output buffer named out */
-#define OUT_LEFT out[0]
+#define OUT_L out[0]
 
 /** shorthand macro for simplifying the reading of the right 
  *  channel of a non-interleaved output buffer named out */
-#define OUT_RIGHT out[1]
+#define OUT_R out[1]
 
 /** shorthand macro for simplifying the reading of the left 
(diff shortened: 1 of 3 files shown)


stephenhensley:
There's some cleanup and such that I'd like to do on the implementation.

I think we may move a few more things out of the class so they're just in the patch_sm namespace, but the current examples have been updated to work with everything as it is now.

So we'll probably do another round of refactor when we we add the next batch of examples.

This board is still pre-release anyway. So only a few people will be affected by the change.

PR #402, opened 2021-09-20
Board/seed1.1

stephenhensley:
Adding runtime check for new rev5 seed (Daisy Seed 1.1)

Pin compatible hardware, with WM8731 codec instead of AK4556.

All DaisySeed firmware should be compatible with the new hardware upon recompiling with this update.

### DIFF
--- a/src/daisy_seed.cpp
+++ b/src/daisy_seed.cpp
@@ -227,16 +227,6 @@ void DaisySeed::ConfigureQspi()
 }
 void DaisySeed::ConfigureAudio()
 {
-    // SAI2 - config
-    // Example Config
-    //      SAI2 Pins (available on pinout)
-    //    pin_group = sai_handle.sai2_pin_config;
-    //    pin_group[DSY_SAI_PIN_MCLK] = dsy_pin(DSY_GPIOA, 1);
-    //    pin_group[DSY_SAI_PIN_FS]   = dsy_pin(DSY_GPIOG, 9);
-    //    pin_group[DSY_SAI_PIN_SCK]  = dsy_pin(DSY_GPIOA, 2);
-    //    pin_group[DSY_SAI_PIN_SIN]  = dsy_pin(DSY_GPIOD, 11);
-    //    pin_group[DSY_SAI_PIN_SOUT] = dsy_pin(DSY_GPIOA, 0);
-
     // SAI1 -- Peripheral
     // Configure
     SaiHandle::Config sai_config;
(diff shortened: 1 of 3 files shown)


src/daisy_seed.h
```diff
@@ -149,6 +149,22 @@ class DaisySeed
     */
     using Log = Logger<LOGGER_INTERNAL>;
 
+    /** Internal indices for DaisySeed-equivalent devices 
+     *  This shouldn't have any effect on user-facing code,
+     *  and only needs to be checked to properly initialize
+     *  the onboard-circuits.
+    */
+    enum BoardVersion
```
TheSlowGrowth, line 157: can we make this an `enum class` - unless casting to int is done regularily, enum class is typically the more strong typed option

src/daisy_seed.cpp
```diff
@@ -281,24 +301,25 @@ void DaisySeed::ConfigureDac()
...
+    pincheck.pin  = {DSY_GPIOD, 3};
+    dsy_gpio_init(&pincheck);
+    if(!dsy_gpio_read(&pincheck))
+    {
+        return DAISY_SEED_1_1;
+    }
+    else
+    {
+        return DAISY_SEED;
+    }
```
TheSlowGrowth, line 324:
```suggestion
    if(!dsy_gpio_read(&pincheck))
        return DAISY_SEED_1_1;
    else
        return DAISY_SEED;
```

src/daisy_seed.cpp
```diff
@@ -245,23 +235,53 @@ void DaisySeed::ConfigureAudio()
...
+            i2c_config.mode           = I2CHandle::Config::Mode::I2C_MASTER;
+            i2c_config.periph         = I2CHandle::Config::Peripheral::I2C_2;
+            i2c_config.speed          = I2CHandle::Config::Speed::I2C_400KHZ;
+            i2c_config.pin_config.scl = {DSY_GPIOH, 4};
+            i2c_config.pin_config.sda = {DSY_GPIOB, 11};
+            I2CHandle i2c_handle;
+            i2c_handle.Init(i2c_config);
+            Wm8731::Config codec_cfg;
+            codec_cfg.Defaults();
+            Wm8731 codec;
```
TheSlowGrowth, line 262: that's a temporary object on the stack. Do we really not need this object anymore after the initial `codec.Init()`? Just being cautious here

TheSlowGrowth (approved):
Ah, the factory burndown implications... Great to see that the Seed will continue to be supported in this form factor.

src/daisy_seed.h
stephenhensley: yeah, actually had the same thought after I pushed. I don't see ever needing to cast as int. So will totally do this.

src/daisy_seed.cpp
stephenhensley: In terms of compatibility with the original Daisy Seed, there aren't any additional functions that were also possible with the 4556. And since `Init` is the only function the current WM8731 driver can do, I don't see a point of keeping it around once its initialized.

That said, I have other boards where I have it as a class member so that when we add more control functionality to the driver (level/filter adjustment, etc.) it can be accessed through the board class.  So I could go either way on this, but in the interest of keeping the Seed -> Seed1.1 transition as seamless as possible, I'm leaning on the side of it not being kept.

TheSlowGrowth: okay, if its not needed after the initial setup, then it can be safely destroyed afterwards

TheSlowGrowth (approved):

PR #403, opened 2021-09-22
fix cmake

TheSlowGrowth:
add missing files to CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,9 @@ add_library(${TARGET} STATIC
     ${MODULE_DIR}/per/spi.cpp
     ${MODULE_DIR}/per/tim.cpp
     ${MODULE_DIR}/per/uart.cpp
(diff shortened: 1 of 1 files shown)


stephenhensley (approved):
This brings up an interesting issue -- without making a dummy program that just loads an instance of every available object, is there a way to catch missing sources like this in CI?

brbrr:
Probably integrating gtest into cmake project is the way to go. But in the end, if we want to maintain two (independent) build systems, it might happen anyway.

TheSlowGrowth:
Problem is that these types of issues come up at link time - so you actually need to USE that class to see the error...

Integrating Gtest into CMake is not as simple as it sounds. Normally that's an easy thing, but in our case Gtest is compiled for x86 while libDaisy is compiled for ARM - and CMake is not designed to handle two different toolchains in the same project...

PR #405, opened 2021-09-26
Use ASFLAGS instead of CFLAGS for .s files.

SSteve:
When I tried to include an assembly file in my project, I was getting errors. It seems like the Makefile is using the wrong flags when building assembly files. When I made this change, my assembly file built and ran correctly. I don't know that this fix is 100% correct but it did work for me. I'm also not sure of the significance of the two different Makefiles so I changed them both.

The sources for my test project are [at this thread](https://forum.electro-smith.com/t/how-do-i-add-an-assembly-file-to-a-project/1969/2) on the Daisy Forum.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -344,7 +344,7 @@ $(BUILD_DIR)/%.o: %.cpp Makefile | $(BUILD_DIR)
 
 $(BUILD_DIR)/%.o: %.s Makefile | $(BUILD_DIR)
 	mkdir -p $(@D)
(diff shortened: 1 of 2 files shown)


stephenhensley:
Good catch! Thanks for the fix!

The only previous assembly file we used within the libraries was the startup file. Not sure why that didn't throw the errors at the time. Possible the Makefile got updated some time after we stopped using that file.

Cool to see someone messing with assembly on the daisy.

PR #411, opened 2021-10-06
Update default blocksize

beserge:
Set to  4 on all plaforms


TheSlowGrowth:
May I politely ask why this change is required?
Strictly speaking, this would actually be a breaking change (resulting in a major version bump) since it silently changes the behaviour of the library for consumers.

PR #413, opened 2021-10-07
Patch SM cvout fix

CorvusPrudens:
For the Patch Submodule, the CV output is converted from a floating point representation of the voltage to the appropriate 12-bit code for the DAC. The initial factor, `819.2`, is `4096 / 5` . A value of 4096 sent to the DAC will result in a voltage output of zero. 

Even with the correct value of `819`, the value should really be clamped since it's not very user friendly for the voltage to wrap as soon as it barely exceeds 5.

### DIFF
--- a/src/daisy_patch_sm.cpp
+++ b/src/daisy_patch_sm.cpp
@@ -227,7 +227,7 @@ namespace patch_sm
         pimpl_ = &patch_sm_hw;
         /** Initialize the MCU and clock tree */
         System::Config syscfg;
-        syscfg.Defaults();
+        syscfg.Boost();
 
         auto memory = System::GetProgramMemoryRegion();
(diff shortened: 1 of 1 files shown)


CorvusPrudens:
[set default clock rate for daisy patch sm to be 480MHz in libDaisy](https://trello.com/c/MqvhPhTd/267-set-default-clock-rate-for-daisy-patch-sm-to-be-480mhz-in-libdaisy)

stephenhensley:
Haha, no wonder doing a simple blink with the DAC wasn't outputting anything when I tried it the other day.

Nice catch. 

Approved both changes. In the future, we'll probably want to add notes about system-level changes (default clock freq) to the changelog, but since the Patch SM is still technically pre-release hardware its fine.

PR #415, opened 2021-10-08
Fix SetLedRaw overflow bug

### LINKED ISSUE #329: PCA9685 LED Driver maximum brightness dips
I noticed this a while ago, and I guess totally forgot to make an issue. Just want to be clear this is not new or related to the minor change in #327 

It seems like when an LED is set to its maximum value it actually dips in brightness a little bit. I first noticed this on the Daisy Field, but I have since noticed it on a few custom boards as well. 

I haven't really had a chance to investigate it yet, but it can be fairly noticeable in certain settings. So I want it to be documented.

I suspect it might be the result of having the on/off times being super close to each other. I don't recall experiencing this issue with old pre-daisy PCA9685 code that  I have, but it's been a while since I've looked at the differences between what we have now and some of my 2-3 year old C drivers.

@TheSlowGrowth have you noticed this behavior on your end as well? I know you've used this led driver quite a bit as well.

beserge:
`off = rawbrightness + on`  
`on = startcycle = ledindex << 2`.
With lots of drivers `ledindex` can get quite large,  so eventually `off = (rawbrightness + on ) > 4095`.

There's an if which sets the fullbrightness flag which I changed to
`rawbrightness + on > 4095` rather than `rawbrightness > 4095`

fixes #329

### DIFF
--- a/src/dev/leddriver.h
+++ b/src/dev/leddriver.h
@@ -240,12 +240,12 @@ class LedDriverPca9685
             draw_buffer_[d].leds[ch].regAddr     = PCA9685_LED0 + 4 * ch;
             draw_buffer_[d].leds[ch].on          = startCycle;
             draw_buffer_[d].leds[ch].off         = startCycle;
-            draw_buffer_[d].updated[ch]          = false;
+            draw_buffer_[d].updated[ch]          = true;
             transmit_buffer_[d].leds[ch].regAddr = PCA9685_LED0 + 4 * ch;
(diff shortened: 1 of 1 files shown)


beserge:
OK I've got the fix in as described earlier. For now I'm keeping track of updates on a per buffer basis rather than per led, which allows us to keep utilizing the DMA to its full extent.

There is a slightly noticeable blink when you push the led over the edge into the next frame, but I believe it's a function of the previously mentioned hardware aspect rather than the software.

PR #418, opened 2021-10-13
Add a driver implementation for the MAX11300

recursinging:
This is a very libDaisy and Eurorack oriented driver for the [MAX11300](https://datasheets.maximintegrated.com/en/ds/MAX11300.pdf) DAC/ADC/GPIO device.  

### MAX11300

> The MAX11300 integrates a PIXI™, 12-bit, multichannel, analog-to-digital converter (ADC) and a 12-bit, multichannel, buffered digital-to-analog converter (DAC) in a single integrated circuit. This device offers 20 mixed-signal high voltage, bipolar ports, which are configurable as an ADC analog input, a DAC analog output, a general-purpose input (GPI), a general-purpose output (GPO), or an analog switch terminal.

The MAX11300 is an interesting device for Eurorack applications not only because of its configurability, but also because of the voltage ranges it can support.  It is possible to power the MAX11300 from the +12V and -12V Eurorack standard power rails, which allows the DAC and ADC to operate at Eurorack voltage levels, without any external scaling circuitry.  This essentially means one could hook up the 20 port pins of the MAX11300 directly (or better perhaps through a small value resistor) to the 3.5mm jacks of a Eurorack module.

This flexibility comes with some limitations.  The configurable voltage ranges are as follows:

* 0V to +10V
* -5V to +5V
* -10V to 0V

...all of these ranges span 10 volts.  At 12 bits of DAC and ADC resolution, the accuracy is probably not good enough for V/Oct use (Although the MAX11300 based [Polyend Poly2 module](https://polyend.com/manuals/poly2-manual/#inputs-outputs) claims better than 5mV accuracy through calibration).  Another limitation of the device DAC is the settling time.  The datasheet recommends allowing for 40us between DAC updates.  In practice I discovered that this time is cumulative per configured DAC pin.  That means the more DAC pins that are configured, the longer the wait between updates, which reduces the time resolution considerably.  The update rate is fine for LFOs and envelopes, but not for audio rate (>100Hz) modulation.

### API

The device can be attached to libDaisy hardware via one of the SPI peripherals.  By default the MAX11300 will use `SPI_1` and its standard pin configuration, but this can be changed by providing your own `SpiHandle::Config`.

```c++
MAX11300 max11300;
MAX11300::Config max11300_config;
max11300_config.Defaults();
max11300.Init(max11300_config);
```

The API of this driver was designed to provide intuitive access to the features most relevant to Eurorack.  
This includes configuring and re-configuring pins at run-time. Pins can be configured in four modes:

* DigitalRead
* DigitalWrite
* AnalogRead
* AnalogWrite

When configuring pins to `DigitalRead` or `DigitalWrite` modes, then an additional configuration `threshold` allows you to set the voltage threshold/output of boolean true.  For example, `8.0f;` would configure a `DigitalWrite` pin to switch between 0V (false) and 8V (true), whereas `2.5f;` would configure a `DigitalRead` pin to return boolean true when the voltage detected at that pin is higher than 2.5V.

```c++
max11300.ConfigurePinAsDigitalWrite(MAX11300::PIN_12, 8.0f);
max11300.ConfigurePinAsDigitalRead(MAX11300::PIN_8, 2.5f);
```

When configured as `DigitalRead` or `DigitalWrite`, pins can be written to and read from as boolean values.

```c++
max11300.WriteDigitalPin(MAX11300::PIN_12, true);
bool value = max11300.ReadDigitalPin(MAX11300::PIN_8);
```

When configuring pins as `AnalogRead`, or `AnalogWrite`, a `VoltageRange` is included to configure what voltage range the pin will read from, or write to.

```cpp
max11300.ConfigurePinAsAnalogRead(MAX11300::PIN_14, MAX11300::VoltageRange::ZERO_TO_10);
max11300.ConfigurePinAsAnalogWrite(MAX11300::PIN_5, MAX11300::VoltageRange::NEGATIVE_5_TO_5);
```

When configured as `AnalogRead`, or `AnalogWrite`, pins can be written to and read from using the raw 12 bit codes, or a by voltage, which is then clamped and scaled according to the voltage range of the configured pin.

```cpp
max11300.WriteAnalogPinVolts(MAX11300::PIN_5, -3.25f);
```

All pin `Read*` and `Write*` method operate locally on memory locations that need to be synchronized with the device.  This is done by invoking the `Update()` method at regular intervals.

```cpp
max11300.Update();
```

### Implementation details

After receiving some helpful advice, I decided to make the driver implementation templated with an abstract "transport".  This allowed for more thorough unit testing, as well as support for a DMA based transfers in the future.  

The (quasi) full-duplex nature of communication with this device required extending the API of the `SpiHandle` class in libDaisy.  I added a `BlockingTransmitAndReceive` method to support sending and receiving data in the context of a single transaction (CS/NSS low). The impact of this addition should be low, as it is simply a wrapper for the underlying HAL pimpl method of the same name.

I discovered that if you try and update DAC configured pins without respecting the datasheet recommended 40us settling time, the MAX11300 will simply lock up until it is power cycled again. In order to prevent this, I added a `wait_us` parameter to the transport `Transmit` interface.  This allows the driver to communicate the max update frequency to the transport implementation - which will be necessary when implementing a DMA based transport.

### Out of scope

The driver transport is currently using blocking methods of the `SpiHandle` class.  This is not optimal, and the nature of the driver API should make a DMA based transport possible.  There are a couple of things to be aware of when tackling this.

* The `SpiHandle` class does not yet have the DMA equivalent to `BlockingTransmitAndReceive`
* The TX and RX buffers are not in the `DMA_BUFFER_MEM_SECTION`, nor does the API allow them to be passed in. This means the DMA TX and RX method impls will need to do cache flushing and invalidation.
* The `wait_us` parameter passed from the driver will need to be respected when scheduling DMA transactions.

### Conclusion

I hope the addition of this driver will open up a number of new options for people designing Eurorack modules on the Daisy platform. I welcome any and all critique and contributions!


recursinging:
Thanks @ohmtech-rdi for the comprehensive review!

There is a lot to discuss, but since I can't get to everything at once, I'll start with the `GPI` behavior I mentioned earlier, since I tested it pretty thoroughly.   

> What happens if you center as well the GPI configuration?

This unfortunately does nothing, and it's clearly stated on page 44 of the datasheet in the description of mode 1 "The input voltage must be between 0V and 5V".  Indeed `GPO` pins can only produce positive voltages as well.

> What happens if you use an ADC instead of a GPI (and do the threshold in software)?
> ...
> I would say let's remove the GPI mode altogether. What do you think?

That's an option, and I'm open to it.  There may be performance implications to consider.  Reading and writing `GPI` and `GPO` pins respectively require a 5 byte SPI transaction, whereas the ANALOG_IN, and ANALOG_OUT transactions are (N*2)+1 bytes, where N is the number of pins configured in those modes.  

In defense of the MAX11300 in this case, by applying a voltage below 0V to a GPI pin, we are using the device outside of its limits (no matter how poorly documented they are). Undefined behavior is to be expected.  

I have more of a problem with having `PinMode` and `VoltageRange` independently configurable, even in invalid combinations - this just a ticket to a foot-gun convention. We all know nobody really reads code docs.

In Java I'd "solve" this with a type system to prevent an invalid configuration.  There's probably a zero-cost, C++ way to do it as well, but I'd need to figure it out first.

recursinging:
While I would have liked to explore the API design that @TheSlowGrowth suggested, I simply don't have the time to implement and adapt the unit tests to it.  The same goes for some of the test refactoring suggestions that @ohmtech-rdi made.  I need to wind all my side projects down by the end of the year, and I'd prefer to invest my time in a DMA transport for this driver.  I've therefore chosen the simpler approach to securing the configuration API that I mentioned in an earlier comment.

Because the DMA transport will require heavy refactoring of the `SpiHandle` class, I'm want to remove it from the scope of this PR, and I'd like to see it merged as-is, with the current blocking transport.

Regardless whether DMA or blocking, I tested this device a bit further and can report the following:

Using: `SpiHandle::Config::BaudPrescaler::PS_2` 

* 41 byte SPI transactions take 26.3 us
* 5b byte SPI transactions take 3.2 us

...as expected, the update rate for GPI/GPO and ADC configured pins reflect this...

* All pins configured GPI/GPO, max update rate - 180 Khz (5.5us period)
* All pins configured ADC, max update rate - 34 Khz (28.5us period)

...the above mentioned limitation on the DAC update rate is painfully visible in this worst-case scenario:

* All configured DAC, max update rate - 1.19 Khz (833us period)

The current transport implementation does not block for the entire 833us period, but instead discards transactions dispatched before the required wait period has passed.  This behavior could prove problematic for some use cases, but I can't think of a better solution.

I've tested the current implementation pretty extensively on real hardware, so as long as there are no more glaring issues with this PR, I consider it ready to merge.

stephenhensley:
@recursinging that sounds totally fine. 

I agree that reworking the SPI for the full-duplex DMA stuff is outside the scope of this PR.

Unless anyone else has caught anything, I agree this is ready to merge.

@recursinging can you add a note about the additional device support in the changelog?  I can merge after that. Thanks!

recursinging:
OK, I've added an entry to the changelog as requested

stephenhensley (approved):

stephenhensley:
Fantastic! -- Approved! Thanks again for the PR!

Electro-Jam:
I wanted to say how much I love that this device works. :)  
I have it in a custom device and it is working great for me.

I wanted to bring this bug to your attention: #549 
When I was putting together my repro and doing a clean build I came across this issue.

PR #425, opened 2021-10-23
Oled screen auto power off

Tom97531:
Power off oled screen after 20s

### DIFF
--- a/src/ui/UI.cpp
+++ b/src/ui/UI.cpp
@@ -39,11 +39,15 @@ UI::~UI()
 
 void UI::Process()
 {
+    uint32_t currentTimeInMs;
+
     // handle user input
     if(!isMuted_)
     {
+        currentTimeInMs = System::GetNow();
         while(!eventQueue_->IsQueueEmpty())
         {
+            lastEventTime_        = currentTimeInMs;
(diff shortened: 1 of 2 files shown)


beserge:
Just checked this out using Torus as a test platform. 

1. When I didn't explicitly set ```screenSaverTimeOut``` it caused a very short timeout which basically made the display flicker on each time you moved a control. Maybe we canb resolve this by having it default to 0.
2. It would be cool if the user could specify their own screensaver animation, although that's very much a thing we can just add later. 

If you look into fixing 1 I'd say this is good to merge from a feature perspective.

stephenhensley:
Ouu, I agree with @beserge for custom screensaver animation, but I think that's out side of the scope of the PR. So I've made a new issue just for that.

Since a few of the other fields in the `UiCanvasDescriptor` will also cause issues when not explicitly set I think it's okay to leave it for now.. 

Once we get a few of the open PRs merged, and the GPIO converted to C++ I want to map out going through and cleaning up the existing inconsistencies within various parts of the API. So we can add a constructor for any logical defaults during that.

I think for now we can merge it.

stephenhensley:
@Tom97531 thanks for the contribution!

PR #429, opened 2021-11-01
Map the SdmmcHandler::Speed config to the GPIO drive strength config

recursinging:
There was [a discussion](https://github.com/electro-smith/libDaisy/pull/311#issuecomment-796048896) in the context of #311 where I mentioned the struggles I faced using the peripheral with the default GPIO speed/drive strength configuration.

Essentially, the peripheral GPIO pins are configured by default with `GPIO_SPEED_FREQ_VERY_HIGH`, which sets the highest drive strength / lowest slew rate for the pin signalling.  On a breadboard, or in the case of poor PCB routing, this can cause ringing significant enough to stop the peripheral from functioning.  Unfortunately, this issue has come up a few times.

By applying the following mapping...

|  Speed  |  GPIO_SPEED  |
|-------------|----------------------|
| LOW       | GPIO_SPEED_FREQ_LOW|
| MEDIUM_SLOW | GPIO_SPEED_FREQ_MEDIUM |
| STANDARD | GPIO_SPEED_FREQ_HIGH |
| FAST | GPIO_SPEED_FREQ_VERY_HIGH|
| VERY_FAST | GPIO_SPEED_FREQ_VERY_HIGH|
 
..we are making the behavior a little more predictable, as lowering the configured speed will increase the slew rate, which in turn should make most breadboard applications work at lower speeds.

I mapped the GPIO speed config to the clock divider here because the config struct is out of scope at the point the GPIO configuration is applied, and I didn't see the need to persist it - I don't expect the clock divisions will change.

### DIFF
--- a/src/per/sdmmc.cpp
+++ b/src/per/sdmmc.cpp
@@ -55,19 +55,31 @@ void HAL_SD_MspInit(SD_HandleTypeDef* sdHandle)
             PC8     ------> SDMMC1_D0 
         */
 
+        // Adjust the gpio drive strength based on the clock divider,
+        // which is derived from the Speed config
+        uint32_t gpioSpeed = GPIO_SPEED_FREQ_VERY_HIGH;
+        switch(sdHandle->Init.ClockDiv)
+        {
+            case 250: gpioSpeed = GPIO_SPEED_FREQ_LOW; break;     // SLOW
+            case 8: gpioSpeed = GPIO_SPEED_FREQ_MEDIUM; break;    // MEDIUM_SLOW
+            case 4: gpioSpeed = GPIO_SPEED_FREQ_HIGH; break;      // STANDARD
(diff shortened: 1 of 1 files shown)


stephenhensley:
Thanks for making the PR -- I recall doing some brief testing with just a single adjustment previously and it caused some issues on existing hardware.

This looks like it should work great, though.

We'll do some testing over here on a few different platforms to make sure it doesn't cause any unexpected issues sometime this week, and get it merged!

stephenhensley:
We did a bit of testing, and everything here seems solid. We didn't have any hardware that was intermittent or breadboard-friendly to specifically see if it fixes the issues reported, but it didn't seem to introduce any issues to existing, functional hardware.

recursinging:
Cool, thanks for bringing it in!

PR #430, opened 2021-11-04
Add MCP23017 I2C driver

brbrr:
This is a libDaisy adaptation of https://github.com/blemasle/arduino-mcp23017 (kudos for such well-documented library @blemasle) for [MCP23017](https://ww1.microchip.com/downloads/en/devicedoc/20001952c.pdf) I/O Expander

The driver supports only the polling mechanism. In my use-case (w 2 MCP chips), I poll 8x6 matrix keyboard  + 4 encoders. CPU utilization increases ~0.5%, which is not much at all!

**API**
It is quite simple:

```
Mcp23017Transport mcp;
mcp.Init();
mcp.PortMode(MCP23017Port::A, 0xFF); // Inputs
mcp.PortMode(MCP23017Port::B, 0xFF);
mcp.Read();
mcp.GetPin(2);
 ```

There are other getters and setters for more precise control.

### CODE REVIEW

**Comment by stephenhensley** [comment]
Very cool to see another new device added to the library!

having "Transport" in the class name here feels a bit misleading since the actual device access, and the communication transport are handled within the class. 

You could probably leave the name of the class as `Mcp23017`.

Though, we may want to have this be a generic `Mcp23X17` where it can be used for either the I2C or SPI versions of the ICs (since afaik they're both functionally identical except for their communication specs).

If that's the case we may want to create the `Mcp23017Transport` (I2C) as a simple transport that just handles the `ReadRegister`/`WriteRegister` flow. (similar to how we have set up a few other recent devices for multiple protocols). 

That would make it easiest for someone else to later add the `Mcp23S17Transport` (SPI) class without having to rework a bulk of how the device is set up, or copy/paste everything to a completely separate class.

If that doesn't seem like you something you want to do, we can review this as-is, and get it merged in soon 😄

**Comment by brbrr** [comment]
Good call! I extracted the I2C transport as you suggested.

**Comment by brbrr** [comment]
@stephenhensley Do you think you can give it another shot?

**Comment by stephenhensley** [comment]
@brbrr yup! we can take another look this week, thanks for the ping.

File: src/dev/mcp23017.h [hunk 1]
```diff
@@ -0,0 +1,401 @@
...
+        //MIRROR = 	0 : use configureInterrupt
+        //SEQOP = 	1 : sequential operation disabled, address pointer does not increment
+        //DISSLW = 	0 : slew rate enabled
+        //HAEN = 	0 : hardware address pin is always enabled on 23017
+        //ODR = 	0 : open drain output
+        //INTPOL = 	0 : interrupt active low
+        WriteReg(MCPRegister::IOCON, 0b00100000);
+
+        //enable all pull up resistors (will be effective for input pins only)
+        WriteReg(MCPRegister::GPPU_A, 0xFF, 0xFF);
```
stephenhensley, line 121: This seems like it should be a part of the Device Initialization, and not a part of the Transport Initialization (since you'd probably want this to happen whether you were using the SPI or I2C versions of the device).

src/dev/mcp23017.h
```diff
@@ -0,0 +1,401 @@
...
+    return static_cast<MCPRegister>(static_cast<uint8_t>(a)
+                                    + static_cast<uint8_t>(b));
+}
+
+enum class MCPMode : uint8_t
+{
+    INPUT,
+    INPUT_PULLUP,
+    OUTPUT,
+};
```
stephenhensley, line 67: I think these can all move inside of the primary device class instead of being global symbols (Unless there's a reason to leave them outside).

If you do move them inside the class we can remove the `MCP` prefix from each of the enum classes (i.e. `Register`, `Port`, `PinMode`) since they'll be scoped to the MCP23x17 class.

stephenhensley (changes requested):
Other than the few organizational things:

1. We should probably rename the file `mcp23x17.h` instead since it's sort of for both versions of the device at this point.
2. Can you add a note to the changelog that you've added a driver for the device.

stephenhensley:
@brbrr totally forgot to actually tag you at any point in the last review comments I made. Just wanted to make sure you saw.

src/dev/mcp23017.h
brbrr: Hmm, I think both `Mcp23017Transport` and `MCP23x17` classes use `MCPRegister` enum. I don't think I know the way to forward declare it. Any ideas?

brbrr:
@stephenhensley No worries. I was a bit side-tracked with other stuff.

src/dev/mcp23017.h
stephenhensley: Oh, gotcha! No worries. I must have missed the shared usage when I was reading through it. 

Totally fine to leave it outside the class in that case.

brbrr:
yeah, it works flawlessly in my custom board (with 2 of these on the single i2c bus)

PR #431, opened 2021-11-05
Bootloader binary incorporation

CorvusPrudens:
This adds the pre-built bootloader binary to libDaisy, along with a couple linkers for programs that want to run from the bootloader. We don't have a convenient mechanism for using those linkers at the moment, but I'd rather get them in place now.


CorvusPrudens:
Ran a quick test to make sure the target works. Looks all good!

CorvusPrudens:
[add bootloader to libDaisy](https://trello.com/c/DTHLWWuf/316-add-bootloader-to-libdaisy)

core/Makefile
```diff
@@ -61,6 +61,8 @@ C_INCLUDES += -include stm32h7xx.h
 #ASM_SOURCES += \
 #$(SYSTEM_FILES_DIR)/startup_stm32h750xx.s
 
+BOOT_BIN = $(wildcard $(SYSTEM_FILES_DIR)/dsy_bootloader*)
```
stephenhensley, line 64: Let's make this conditionally assigned with `?=` instead.

That'll make testing new version of the bootloader, or relying on older versions of the bootloader a bit easier in the future.

stephenhensley (changes requested):
Looks good other than the one var in the core/Makefile.

stephenhensley (approved):

PR #433, opened 2021-11-05
USB host interface

CorvusPrudens:
This PR depends on  #432 and adds the actual interface for handling USB host interactions.


stephenhensley:
As a basic wrapper around the STM32 middleware this seems to work fine.

I do want us to have the Config (even though its empty for now) passed into the `Init` function of the hid/msd class.

That way we can add stuff to that (user callbacks, etc. for #444).

PR #438, opened 2021-11-20
feat: add 44.1kHz to SAI

TheSlowGrowth:
Is this all that's needed to add support for 44.1kHz? It seems too simple, so I guess I'm missing something.

44.1kHz makes sense when sample playback from *.wav files is desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 * bootloader: the bootloader can now be flashed directly from libDaisy using `make program-boot`
 * driver: added support for the MAX11300 ADC/DAC/GPI/GPO device
+* SAI: add support for 44.1kHz sample rates
+* patch_sm: add support for 44.1kHz sample rates
 
 ### Bug fixes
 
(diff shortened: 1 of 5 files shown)


stephenhensley:
Hahahaha, if only. Right?

Because of the SAI clock source this will still end up generating a 48kHz clock. 

In order to add 44.1kHz (and the other related samplerates) we'll have to adjust the PLL for the clock.

I can't recall if the PLL (2?) is used exclusively for audio/ADCs at this point, but that's the goal. And then we can adjust the PLL/MCLK freq to target the desired 256fs speed.

Once something like that is set up, we could probably implement something that can hit pretty much any target samplerate with good accuracy.

stephenhensley:
Hi @TheSlowGrowth 

I'm starting a little bit of housekeeping, and just want to tidy up some stale PRs.

I've marked this as a draft for now. If you'd rather close it for the time being instead, that's also fine. It'll probably still be a little bit before we get around to reworking the clock tree, adding an API for adjusting the PLLs for the required frequency changes on our end unless someone else gets to it sooner.
Thanks!

PR #434, opened 2021-11-05
USB MSD FatFS 

CorvusPrudens:
This PR depends on #433 and brings in the FatFS code for the MSD class to actually talk to USB drives.

I'll need to actually test this out to make sure it build and functions correctly, but that should only take a few minutes.

PR #448, opened 2021-12-02
FatFS Integration for USB and/or SDMMC

stephenhensley:
Builds off of the work done by @CorvusPrudens in #434 

Made a few changes:

* Unified the two separate interfaces (usbh_fatfs, and fatfs) to a single interface that can initialize either/both volumes
* Added cache maintenance to usb_diskio
* Fixed up WavPlayer class to be agnostic to the hardware/media since it was doing it's own FATFS init internally that wouldn't work well with the new setup. (There's still something kind of funky going on with it that seems to only be an issue with the SD Card).
*  Increased FATFS _VOLUMES to 2 to support multi-volume mounting of SD card and USB drive simultaneously

For a usage for SD and/or USB example:

```cpp
    /** Init the SD card you want to use */
    SdmmcHandler::Config sd_cfg;
    sd_cfg.Defaults();
    sdcard.Init(sd_cfg);

    /** Init USB drive */
    MSDHandle::Config usbcfg;
    usb.Init(usbcfg);
    /** And wait until ready -- this could/should probably turn into a callback? */
    while(!usb.GetReady())
        usb.Process();

    /** Link Daisy to FatFS configured for SD card 
     *  Once this is done, FatFS can be used as usual. */
    FatFSInterface::Config fsi_cfg;
    fsi_cfg.media
        = FatFSInterface::Config::MEDIA_USB | FatFSInterface::Config::MEDIA_SD;
    fs_inf.Init(fsi_cfg);

    /** Get the path to the root directory we want to use:
     *  0:/ is the first, registered path
     *  1:/ is the second registered path .
     * 
     *  volume registration occurs in the order presented in the FatFSInterface::Config::Media enum
     *  i.e. 0:/ is the SD card , and 1:/ is the USB drive if both are used.
     * 
     *  When only one volume is used it will always be at 0:/
     */
    const char* rootpath = fs_inf.GetUSBPath();

    FATFS& fs = fs_inf.GetUSBFileSystem();

    /** Mount the FATFS filesystem */
    f_mount(&fs, rootpath, 1);

```

I don't necessarily like that the user has to manage all three pieces on their own, but with the ability to configure the SDMMC hardware, I'm not sure there's a better solution.

I'd also like to avoid providing wrappers for all of fatfs since it's well documented on it's own, and very easy to work with.

Open to ideas on simplification, and improvement, but this works.


stephenhensley:
TODO:

* [ ] Potentially change name of MSDHandle to USBHostHandle or similar (all host classes can be enumerated through the single host class, and we can gradually add support for the other classes over time).
* [ ] Add note in documentation that FatFSInterface can't be placed in DTCMRAM (default location of the stack) since this is off-limits for SDMMC1 and not-DMA accessible for USB periph. (actually for SDMMC the only valid location is in AXI SRAM).

PR #451, opened 2021-12-06
C++ GPIO

stephenhensley:
This is essentially a duplicate of the work done in #354 except that it preserves the C-style GPIO API. Idea being that we'll add planned deprecation of the current usage without breaking any code. Some time next year we'll remove the old C style, but this will allow for a gradual replacement.

We'll need to go through and adapt the rest of the library to use the new `Pin` type instead of the old, `dsy_gpio_pin`, but it doesn't have to happen right away. We can do that as gradually as we need to. I've added an operator for a conversion to the old type. So the new struct can already be used anywhere that the old one was used without any other changes necessary.

List of changes:
* Add new `Pin` structure to replace old c-style `dsy_gpio_pin`
* Add new GPIO class with C++ API, and multiple initialization options.
* Added comments to all old dsy_gpio_pin functions noting that code should migrate to using the new `Pin` struct.
* Added `daisy::seed` namespace containing `constexpr Pin` mappings for the entire SOM.
  * A namespace was used instead of having these be DaisySeed members to allow for simplistic, Arduino-like syntax when using the pins. (i.e. `a_gpio.Init(D0, GPIO::Mode::OUTPUT);`
  * There was a brief discussion about putting the DaisySeed inside of this namespace, but this would cause a breaking change. So we will put off doing any namespace reorganization of existing board-classes, etc. until a later time.
* Somewhat out of the scope of the PR: Moved contents of `DaisySeed::Configure` to the `DaisySeed::Init` function, making the `Configure` an empty function. The function was left in place for now to prevent any breaking changes.

These changes stage a few things for future removal. At some point likely Q2-Q3 of 2022 we'll do more substantial update that actually removes these old mechanics, and cleans up the library a bit.

For now we're trying to improve accessibility, and ease of use without breaking existing code.


stephenhensley:
Found something pretty funky: `GPIOPort::PI` is a big red flag because... well.. it's PI.. 🥧 

So we'll loop back around on the naming -- leading contender for the port names now is `PORTA`, `PORTI`, etc.

PR #463, opened 2021-12-15
Bug Fix - USB Logger missing messages with wait_for_pc set to true  

### LINKED ISSUE #343: USB Logger can fail/cause freezing if interrupts start too soon after StartLog
Seems like this is probably a callback function from the USB getting interrupted by the audio IRQs which are higher priority. Specifically I'm guessing the descriptor functions are getting interrupted, but I haven't stepped through everything myself yet.

A workaround for now is to use a delay (500ms seems acceptable) between `StartLog` and any subsequent Init or Start functions that would cause periodic interrupts.

Longterm we should add `ScopedIrqBlockers` to the functions that can cause issues when interrupted.

stephenhensley:
Applied short delay to Logger<>::StartLog function to prevent missed messages.

I can't really explain what wasn't working because the blocking transmit was blocking (I confirmed by stepping into some of the functions), but for some reason without a delay, the first several messages wouldn't ever make it to a Serial Monitor.

Also, trying to put the delay within the USB init instead of the Logger, or pretty much anywhere else that would have made sense failed to stop the issue.

Anyway, this seems to resolve #343  and related issues that required setting a delay after the Logging function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+* logger: Added 10ms delay at the end of `StartLog` function. Without this, messages immediatly following the `StartLog` function were getting missed when `wait_for_pc` is set to `true`.
+
 ### Other
 
 ### Migrating
--- a/src/hid/logger.cpp
+++ b/src/hid/logger.cpp
@@ -3,6 +3,7 @@
 #include <cstdio>
 #include <cassert>
 #include "logger.h"
+#include "sys/system.h"
 
 namespace daisy
 {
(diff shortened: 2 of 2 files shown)


stephenhensley:
Confirmed that this should fix #343 

The following code, doesn't cause any issues starting the audio immediately after the Logger:

```cpp
#include "daisy_seed.h"

using namespace daisy;

DaisySeed hw;

void AudioCallback(AudioHandle::InputBuffer  in,
                   size_t                    size)
{
    std::copy(&in[0][0], &in[0][size], &out[0][0]);
    std::copy(&in[1][0], &in[1][size], &out[1][0]);
}

int main(void)
{
    hw.Init();
    hw.StartLog();
    hw.StartAudio(AudioCallback);
    hw.PrintLine("Hello World!");
    while(1) {}
}
```

PR #464, opened 2021-12-21
Updated Reference Documentation

stephenhensley:
This ended up being a bit bigger than I expected, but the results are certainly worth it.

I'm still going through doing some reorganization of the sections, and some of the new "guides", but it's pretty much wrapped up at this point.

The big changes:

* Custom homepage for documentation that isn't just a copy of the repo README.
* Much prettier docs in general (thanks to the [doxygen-awesome css](https://github.com/jothepro/doxygen-awesome-css) template)
* Lot's of customization of doxygen output to make for a better layout, and easier navigation
* Fixed usage of `ingroup` `defgroup` and `addtogroup` in various places to actual include expected documentation with classes, etc. (previously many things were getting stripped out for some reason for example, the detailed description of the `AbstractMenu` class was just not present in the doxygen output, etc.).
* Added local examples to libDaisy to serve as examples in documentation, as well as providing better coverage via CI
* Added the local examples to the CI workflow
*  Moved the "detailed description" to the top 
* Added a number of getting started guides for ADCs, GPIO, Serial Printing, Audio, and more. These will continue to be refined, and more guides will be written soon as well.
* Moved all of the predefined doxygen groups out of the `daisy.h` file, and into a doc specific `index.h`, and cleaned it up a bit.
* Changed setting `EXTRACT_ALL` to true. This has a few repercussions:
  * All undocumented classes, functions, etc. are now included in the docs (although empty)
  * This in combination with a ton of the small edits throughout the files, resolved all but two warnings when generating the docs (reduced from several hundred 😄 )

The generated: `daisy::GPIO` page for the new C++ GPIO API is going to be the model for new documentation moving forward. This means fully documented functions, initialization information, and an actual example. We'll work through the rest of the library early in 2022 to bring everything up to this new standard.


stephenhensley:
Here's a tentative list of the few things left to do before I think this will be good to merge.

* [ ] Fix examples build CI (currently doesn't have libdaisy.a from previous job)
* [ ] Fix issue with footer / treeview getting blocked
  * This seems to be an incompatibility with some of the manual tweaks I made before introducing the custom css layout
* [ ] Fill in Getting Started - MIDI doc
* [ ] The horizontal rule in the side navbar, and the main body of the page are barely misaligned an it's an eyesore.


I'll probably do a bit more reorganization, or renaming of groups, but a lot of that can happen more atomically later

doc/md/_a3_Getting-Started-Audio.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+as opposed to the non-interleaved version of the same data, that could be represented as:
+
+```
+{ 
+    { L0, L1, L2, L3 },
+    { R0, R1, R2, R3 }
+}
+```
```
TheSlowGrowth, line 199: Maybe switch the order?

> Above, we looked at the non-interleaved format, where each channel had its own array of samples, like this:
> ```
> { 
>     { L0, L1, L2, L3 },
>     { R0, R1, R2, R3 }
> }
> ```
> 
> Interleaved audio is instead one large array containing multiple channels of data. This data is interleaved so that there is a single sample of each channel before moving to the next sample. For example, with a block size of four, the full buffer would look like:
> 
> ```
> { L0, R0, L1, R1, L2, R2, L3, R3 }
> ```

doc/md/_a3_Getting-Started-Audio.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Changing Audio Settings
+
+The AudioHandle allows the user to change certain settings, in particular the:
+
+* Sample Rate
+* Block Size
+
+On the DaisySeed, or other official boards you can use the following functions:
```
TheSlowGrowth, line 208: Please add where in the setup process this needs to be called. I don't think this can be called while audio is running, right?

doc/md/_a3_Getting-Started-Audio.md
```diff
@@ -0,0 +1,253 @@
...
+hw.SetAudioSampleRate(SaiHandle::Config::SampleRate::SAI_48KHZ);
+```
+
+Block size can be any number up to 256.
+
+See the [SaiHandle::Config::SampleRate](@ref daisy::SaiHandle::Config::SampleRate) enum for the available samplerate options. It's also worth mentioning that, depending on what codec is connected, not all options are compatible.
+
+## What Not to do in the Audio Callback
+
+The audio callback is in what's called and "Interrupt Handler" without going in depth on what that means here, it does mean that it the function _must_ take less time than it takes to transfer out the audio buffer, otherwise your audio will start to have under-run errors, which cause digital artifacts in the audio path.
```
TheSlowGrowth, line 220:
```suggestion
The audio callback is in what's called and "Interrupt Handler" without going in depth on what that means here, it does mean that it the function _must_ take less time than it takes to transfer out the audio buffer, otherwise your processed audio will not be ready by the time the hardware need fresh samples to output. This is called "under-run error" and it causes digital artifacts in the audio path.
```

doc/md/_a3_Getting-Started-Audio.md
```diff
@@ -0,0 +1,253 @@
...
+
+The amount of time can be calculated as `(1 / sample_rate) * block_size`. For example, with 48 sample block size at 48kHz, the function can take 1ms before underruns start to occur.
+
+Now, that is a bit of a simplification, because chances are you'll be doing a few things in the main `while()` loop, but it does mean that you shouldn't do things that take an indeterminant amount of time.
+
+These things include:
+
+* calls to other peripheral blocking functions (i.e. DAC Write, SPI tranfer, etc.)
+* dynamic allocation like `malloc`
```
TheSlowGrowth, line 229:
```suggestion
* dynamic allocation like `malloc`
* printing debug information to a display or serial connection
```

doc/md/_a3_Getting-Started-Audio.md
```diff
@@ -0,0 +1,253 @@
...
+{
+    
+    if (action_flag) {
+        // Do the big thing that can take a while
+        initiate_big_transfer();
+        // and clear the flag
+        action_flag = false;
+    }
+}
+```
```
TheSlowGrowth, line 253: This could be added:

## How to measure the processing load

It is crucial to have a feeling of how much processing load your algorithm introduces and how much room for additions you still have. This can be done by measuring the execution time of the audio callback and comparing it to the available time. Daisy comes with a helper class to do this, the `CpuLoadMeter`.

To measure the CPU load, you would call `.OnBlockStart()` at the beginning, and `.OnBlockEnd()` at the end of your audio callback. The you can request the resulting load measurement and print it on a display or a serial connection. It is crucial to do the printing from the main loop, not the audio callback itself, as discussed before.

A complete example would look like this:

```cpp
#include "daisy_seed.h"
#include "daisysp.h"
using namespace daisy;
DaisySeed hw;
CpuLoadMeter loadMeter;

void MyCallback(AudioHandle::InputBuffer in, 
                size_t size) 
{
    loadMeter.OnBlockStart();
    for (size_t i = 0; i < size; i++)
    {
        // add your processing here
        out[0][i] = 0.0f;
        out[1][i] = 0.0f;
    }
    loadMeter.OnBlockEnd();
}
int main(void)
{
    hw.Init();
    
    // start logging to the serial connection
    hw.StartLog();
    
    // initialize the load meter so that it knows what time is available for the processing:
    loadMeter.Init(hw.AudioSampleRate(), hw.AudioBlockSize());
    
    // start the audio processing callback
    hw.StartAudio(MyCallback);
    
    while(1) {
        // get the current load (smoothed value and peak values)
        const float avgLoad = cpuLoadMeter.GetAvgCpuLoad();
        const float maxLoad = cpuLoadMeter.GetMaxCpuLoad();
        const float minLoad = cpuLoadMeter.GetMinCpuLoad();
        // print it to the serial connection (as percentages)
        hw.PrintLine("Processing Load %:");
        hw.PrintLine("Max: " FLT_FMT3, FLT_VAR3(maxLoad * 100.0f));
        hw.PrintLine("Avg: " FLT_FMT3, FLT_VAR3(avgLoad * 100.0f));
        hw.PrintLine("Min: " FLT_FMT3, FLT_VAR3(minLoad * 100.0f));
        // don't spam the serial connection too much
        System::Delay(500);
    }
}
```

doc/md/_a4_Getting-Started-ADCs.md
```diff
@@ -0,0 +1,173 @@
...
+// Initialize the first one connected to A0
+my_adc_config[0].InitSingle(A0);
+// Initialize the second one connected to A4
+my_adc_config[1].InitSingle(A4);
+```
+
+And finally when we go to initialize the ADC itself, we can now omit the `&` that we had to use before because it's an array of objects, and we want to make sure to update the total count that we passed in before:
+
+```cpp
+hw.adc.Init(my_adc_config, 2);
```
TheSlowGrowth, line 155:
```suggestion
hw.adc.Init(my_adc_config, num_adc_channels);
```

doc/md/_a4_Getting-Started-ADCs.md
```diff
@@ -0,0 +1,173 @@
...
+
+Now we've set up a single pot. So let's look at what it would look like to hook up two (or more).
+
+Nearly everything will stay the same except instead of using a single `AdcChannelConfig`, we'll want to create one for each control we want to use. C and C++ have a mechanism for dealing with this known as an array. If you're not familiar, it is basically just an ordered list.
+
+So for setting up two pots we'll change the initialization code to look a little more like:
+
+```cpp
+// Create an array of two AdcChannelConfig objects
+AdcChannelConfig my_adc_config[2];
```
TheSlowGrowth, line 145:
```suggestion
const int num_adc_channels = 2;
AdcChannelConfig my_adc_config[num_adc_channels];
```

doc/md/_a4_Getting-Started-ADCs.md
```diff
@@ -0,0 +1,173 @@
...
+And finally when we go to initialize the ADC itself, we can now omit the `&` that we had to use before because it's an array of objects, and we want to make sure to update the total count that we passed in before:
+
+```cpp
+hw.adc.Init(my_adc_config, 2);
+```
+
+Now you can read from either input. If you want to read the pot connected to A0 you would use `hw.adc.Get(0)`, or `hw.adc.GetFloat(0)`, and if you want to read from the pot connected to A4, you would use `hw.adc.Get(1)`, or `hw.adc.GetFloat(1)`.
+
+You'll notice that you access them by the order in which you initialized them, not by the number associated with the Pin.
+
```
TheSlowGrowth, line 161: ## Order to chaos - naming ADC channels

Once you have a lot of ADC channels, it can be difficult to remember which channel corresponds to which control. To solve this, you can use `enum`s to assign names to channels like this:
```cpp
enum AdcChannel {
   pitchKnob = 0,
   pitchCv,
   gainKnob,
   gainCv,
   NUM_ADC_CHANNELS
}
```

Here, the first value is assigned to `0`and the others increment automatically. Interestingly, this logic can be used to automatically get the total number of channels you use, by adding one last value `NUM_ADC_CHANNELS`. Of course you could also specify the individual values by hand, just add `myKnob = 3,`.

Now you can make your code much easier to read - and maintain, should you decide to change the hardwre connections later:
```cpp
AdcChannelConfig my_adc_config[NUM_ADC_CHANNELS];
// ...
my_adc_config[pitchCv].InitSingle(A0);
// ...
hw.adc.Init(my_adc_config, NUM_ADC_CHANNELS);
// ...
const pitchCvValue = hw.adc.GetFloat(pitchCv);
```

doc/md/_a6_Getting-Started-External-SDRAM.md
```diff
@@ -0,0 +1,74 @@
...
+float __attribute__(section((".sdram_bss"))) my_buffer[1024];
+```
+
+Now, that is a lot of stuff to type, and pretty hard to remember. So we made a shorthand macro to make it a bit easier:
+
+```cpp
+float DSY_SDRAM_BSS my_buffer[1024];
+```
+
+The first, longform example is the spelled out gcc attribute that says what memory section to put the array in. This can be useful on big projects when using one of the many other available memory sections within the STM32H7 processor.
```
TheSlowGrowth, line 39:
```suggestion
The first, longform example is the spelled out compiler attribute that says what memory section to put the array in. This can be useful on big projects when using one of the many other available memory sections within the STM32H7 processor.
```

TheSlowGrowth (changes requested):
These are some great changes that will make it much easier for people to get started! 
I have only some small additions / requests.

stephenhensley:
@TheSlowGrowth Thanks for the thorough review, and great suggestions on additional content! Very helpful!

doc/md/_a1_Getting-Started-GPIO.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Speed Argument
+
+The final argument, `Speed sp` is a special configuration for Output modes that control the slew rate, or the speed at which it can switch from 0V to 3V. We won't go super in depth on that in this, but you'll always want it to be the lowest speed possible for a given application.
+
+### Defaults
+
+So in our example above, we can avoid having to connect any external resistors to our button by using the pull up feature.
+
+If we wanted to use some external resistors we could pull one side high with a resistor to 5V, and the other with a resistor to GND, and connect one side to the GPIO input, and if we didn't need to use the GPIO pull up resistor we could initialize the pin using the default settings:
```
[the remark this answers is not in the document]
stephenhensley: This is a good point, and would probably benefit from a diagram instead of text.

Also, this was mentioned in a previous review, and I forgot to address it, but since the defaults are discussed on a per-parameter basis, this defaults section feels a bit redundant. 

So I'll probably just remove it.

Early next year we're going to go through and add diagrams for a lot of the hardware descriptions throughout these guides.

doc/md/_a1_Getting-Started-GPIO.md
```diff
@@ -0,0 +1,222 @@
...
+
+This argument will default to `Pull::NOPULL` if you only supply a `Pin` and a `Mode`
+
+### Speed Argument
+
+The final argument, `Speed sp` is a special configuration for Output modes that control the slew rate, or the speed at which it can switch from 0V to 3V. We won't go super in depth on that in this, but you'll always want it to be the lowest speed possible for a given application.
+
+### Defaults
+
+So in our example above, we can avoid having to connect any external resistors to our button by using the pull up feature.
```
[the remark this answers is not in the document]
stephenhensley: I've added a little to the Pull section to discuss this -- This is definitely another thing that will become more clear with diagrams.

doc/md/_a1_Getting-Started-GPIO.md
```diff
@@ -0,0 +1,222 @@
...
+
+If we wanted to use some external resistors we could pull one side high with a resistor to 5V, and the other with a resistor to GND, and connect one side to the GPIO input, and if we didn't need to use the GPIO pull up resistor we could initialize the pin using the default settings:
+
+```cpp
+my_button.Init(D0)
+```
+
+By specifying the input mode, and pullup configuration we can do the same thing with just a button, and no external resistors by adding the Mode and Pull arguments:
```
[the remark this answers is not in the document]
stephenhensley: This was in the section I removed, but I do plan on adding some sort of code conventions doc that will cover things like why we use enum class instead of macros, and other language mechanics common to libDaisy -- this can cover optional arguments a bit more in depth.

doc/md/_a3_Getting-Started-Audio.md
stephenhensley: Excellent suggestion! Added.

doc/md/_a4_Getting-Started-ADCs.md
stephenhensley: Another great suggestion. Added as well!

PR #467, opened 2021-12-23
Add ability to compile .cc

beserge:
Add CC_SOURCES and .cc compilation to core/Makefile

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -51,6 +51,7 @@ SYSTEM_FILES_DIR ?= \
 ######################################
 
 CPP_SOURCES ?= 
+CC_SOURCES ?= 
 
 C_SOURCES += \
 $(SYSTEM_FILES_DIR)/startup_stm32h750xx.c
@@ -224,6 +225,8 @@ OBJECTS = $(addprefix $(BUILD_DIR)/,$(notdir $(C_SOURCES:.c=.o)))
 vpath %.c $(sort $(dir $(C_SOURCES)))
 OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(CPP_SOURCES:.cpp=.o)))
 vpath %.cpp $(sort $(dir $(CPP_SOURCES)))
+OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(CC_SOURCES:.cc=.o)))
+vpath %.cc $(sort $(dir $(CC_SOURCES)))
 # list of ASM program objects
 OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(ASM_SOURCES:.s=.o)))
 vpath %.s $(sort $(dir $(ASM_SOURCES)))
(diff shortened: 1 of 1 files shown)


TheSlowGrowth:
And at some point, we add support for *.c++ files endings? 
Hmmm...
Wouldn't it make sense to specify the file ending directly in the CPP_SOURCES list instead?

stephenhensley:
@TheSlowGrowth I was kind of thinking the same thing.

It'd be nice if we could just append to the `CPP_SOURCES` and just have a list of acceptable suffixes. 

I think @beserge and I both tried quick hacks on the existing recipes to do that, but it didn't work quite right. Admittedly, my `make` chops aren't the best.

Would something as simple as this replacement of the separate `CC_SOURCES` list work as expected? 

```
OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(CPP_SOURCES:.cpp=.o)))
vpath %.cpp $(sort $(dir $(CPP_SOURCES)))
OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(CPP_SOURCES:.cc=.o)))
vpath %.cc $(sort $(dir $(CPP_SOURCES)))
```

ndonald2:
@stephenhensley I did a little poking. Your exact proposal does not work as-is because the current `OBJECTS` append invocations for actually pass through file names in the relevant source lists that don't end with the correct extension.

As in, `$(CPP_SOURCES:.cpp=.o)` will reassign `.cpp` extensions in the list to `.o` but leave non-matching extension unchanged.

However this works (filter `CPP_SOURCES` into two sub lists, each containing exclusively `.cpp` or `.cc` files)

```make
CPP_SOURCES_CPP = $(filter %.cpp, $(CPP_SOURCES))
OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(CPP_SOURCES_CPP:.cpp=.o)))
vpath %.cpp $(sort $(dir $(CPP_SOURCES)))
CPP_SOURCES_CC = $(filter %.cc, $(CPP_SOURCES))
OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(CPP_SOURCES_CC:.cc=.o)))
vpath %.cc $(sort $(dir $(CPP_SOURCES_CC)))
```

This is actually a little more foolproof in general because it ensures that no files with incompatible extensions make it through to `OBJECTS`. In fact it might be good to add a similar filter statement to the `C_SOURCES` object append invocation as well.

HeckHeckHeckHeck:
Hya :) good to see i am not first to want this.
This would be solved in my PR: https://github.com/electro-smith/libDaisy/pull/647
sorry i havent seen this PR earlier @TheSlowGrowth. What do you think of my PR, suggestions welcome.
If ppl drive me nuts discussin on end and not progressing i usually just hard fork then especially MIT/BSD style code, but in general i do like to collaborate, so lets try it. 
Cheers

PR #472, opened 2022-01-06
Multi Slave SPI & Multi-Chip MAX11300

TheSlowGrowth:
Sorry this is a rather lengthy PR. Let me know if you'd prefer it split up in several parts

### Breaking Changes

* driver: MAX11300 driver interface changed considerably
* spi: Order of arguments of the `SpiHandle` DMA functions changed
* dma: `SpiHandle` previously used `DMA1_Stream7`, now uses `DMA2_Stream2` and `DMA2_Stream3`

### Features

* spi: `SpiHandle` now has callbacks before and after a DMA transaction starts (can be used for software driven chip select)
* spi: added `MultiSlaveSpiHandle` that allows to connect to multiple SPI slave devices on a shared bus
* driver: MAX11300 now supports multiple chips on a shared bus
* driver: MAX11300 now uses DMA to handle updates without blocking
* driver: MAX11300 now has an auto-update mode where it continuously updates itself
* driver: MAX11300 can now call a user-provided callback after an update is complete

### Bug Fixes

* testing: debugging configuration now uses `lldb` debugging extension to support unit test debugging on macOS with Apple Silicon

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,6 +2,7 @@
 	"recommendations": [
         "ms-vscode.cpptools",
 		"xaver.clang-format",
-        "davidschuldenfrei.gtest-adapter"
+        "davidschuldenfrei.gtest-adapter",
+        "vadimcn.vscode-lldb"
 	]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,6 +15,7 @@
             },
             "preLaunchTask": "build-libDaisy-tests",
             "osx": {
+                "type": "lldb",
                 "MIMode": "lldb",
             },
             "windows": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ SpiHandle::Result DmaTransmit(uint8_t*                            buff,
 // Old: MAX11300 takes no template arguments
 MAX11300 max11300driver;
 // New: MAX11300 takes number of chips as template argument
-const size_t num_devices = 4;
+constexpr size_t num_devices = 4;
 MAX11300<num_devices> max11300driver;
 
 // Old: constructor only takes a config struct
@@ -78,8 +78,7 @@ max11300driver.Update(); // blocks
 //      - Update() can retrigger itself automatically
 //      - A callback can be called after each update cycle that was completed successfully
 void MyUpdateCompleteCallback(void* context) {
-    // use the context to trace back where the callback came from
-    // or ignore it.
+    // The context is the pointer you passed when calling `.Update()`
     // This callback comes from an interrupt, keep is fast.
 }
 max11300driver.Update(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Features
 
 * spi: `SpiHandle` now has callbacks before and after a DMA transaction starts (can be used for software driven chip select)
+* spi: `SpiHandle` now supports simultaneous transmit and receive with DMA using `SpiHandle::DmaTransmitAndReceive()`
 * spi: added `MultiSlaveSpiHandle` that allows to connect to multiple SPI slave devices on a shared bus
 * driver: MAX11300 now supports multiple chips on a shared bus
 * driver: MAX11300 now uses DMA to handle updates without blocking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 * spi: added `MultiSlaveSpiHandle` that allows to connect to multiple SPI slave devices on a shared bus
 * driver: MAX11300 now supports multiple chips on a shared bus
 * driver: MAX11300 now uses DMA to handle updates without blocking
-* driver: MAX11300 now has an auto-update mode where it continuously updates itself
+* driver: MAX11300 now updates the chips continuously until manually stopped
 * driver: MAX11300 can now call a user-provided callback after an update is complete
 
 ### Bug Fixes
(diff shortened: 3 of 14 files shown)


TheSlowGrowth:
I added some small documentation fixups that I found while reading the PR diff this morning. I don't expect more changes from my side - please review @recursinging @stephenhensley @beserge

stephenhensley:
Nice! Thanks for the PR.

We'll review sometime in the next week, and get you any feedback! 

It does look a bit hefty, but I think it's OK to leave it all together.

Happy New Year 🚀

recursinging:
First off, this is an awesome PR. Thanks @TheSlowGrowth for your fantastic work as usual.

I was looking through the changes and noticed there are no provisions for suppressing DAC update transactions for the minimum required wait time of `dac_pin_count_` * 40µS.  I was curious so I tested it, and sure enough, the device locks up if there are more than two pins configured in DAC mode.

I'll try and find some time this week to have a more thorough look through your changes, and see if there is a simple solution.

TheSlowGrowth:
Thanks @stephenhensley , I'll add a note about the improved hardfault handler. It came up when I tested the "flush the cache insteat of using deidcated DAM buffers"-approach. I had some hard faults and ultimately used this code to debug them.
BTW, flushing the cache only works partially: The problem is that you'll have to use a buffer that is NOT on the stack, since the stack sits in a memory region not accessible to the DMA (triggering a hardfault). I guess we should eventually add some checks to the DMA functions just to verify that the buffers are in accessible memory ranges.

Regarding the device-lockup:
For the immediate update mode of the DAC pins, there is a mention of 80µs in the datasheet. This makes sense, in order to update the DAC, the device needs to jump to another pin in its update cycle, and that takes time. In the sequential mode, there is no mention of a waiting time in the datasheet. I would also be surprised to see such a requirement - isn't the idea of the sequential mode that you can write all DAC values in one go using DMA?

I'll investigate further.

recursinging:
So upon further investigation, it seems the root cause of device lockups I have experienced lie elsewhere. At the moment I can verify that DAC pins behave as expected, when using the new cyclic AutoUpdate feature. The behavior I experienced only occurs when invoking Update() from a program loop.  This suggests that there may be issues with the DMA transaction scheduling, which is starting SPI transactions at intervals that the MAX11300 doesn't like.

This all warrants further investigation and I'll report back when I know more.

TheSlowGrowth:
I removed the manual update mode from the driver, for two reasons:
- It didn't actually make much sense. Like the ADC driver, automatic updates are the most useful way of interfacing with the I/O - if you really wanted to update just once, you could call `Start()` followed by `Stop()`
- The manual update mode introduced the bug observed by @recursinging  - it is now gone, according to my testing

In addition, I added the changelog notes for the hardfault handler.

recursinging:
As far as I'm concerned, this is complete.  I tested against real hardware, and I find the `Start()` / `Stop()` strategy to be perfect for this device.  The only remaining issue I have is the DMA Buffer needing to be passed in... It's a leaky abstraction and I really wish there was way to handle it more elegantly.  The discussion and solution is outside the scope of this issue though, so let's get this one merged as-is.

stephenhensley:
Wonderful! Should be able to take a final look, and merge this today or tomorrow!

The `Start()`/`Stop()` for one-shot usage seems fine to me as well.

stephenhensley:
@TheSlowGrowth So I think everything's good! 

The only thing that stood out is that the new `spiMultiSlave.h` isn't included directly in `daisy.h`. However, it is included through the MAX11300 driver. So it is getting included, just indirectly. 

Not sure if it's critical to change or not (and, generally speaking, what the policy on this should be), but otherwise I approve!

TheSlowGrowth:
Ahh, good catch, I'll update that

stephenhensley (approved):

stephenhensley:
Approved!

Thanks again @TheSlowGrowth for the wonderful new features, and @recursinging for the feedback and reviews. You both rock! 🚀

GR8Instruments:
Hi, any new samples to show all of this working that I can look at for my project using the same device?

GR8Instruments:
@TheSlowGrowth, I was looking for a set of samples that shows some of this setup in action.  I was using the older version of the max driver and since I updated to the new libraries, I am looking for a good sample set to both get mine back up and running plus look at how to use the new callback features.  Anything you can point me to would be greatly appreciated.

TheSlowGrowth:
I have it working in a personal project to collect UI hardwdare input (buttons and potentiometers). I use the update complete callback. You can find the code [here](https://github.com/TheSlowGrowth/TapeLooper/blob/feature/TSG/firmware-ui/firmware/src/hardware/UiHardware.h#L219=).
Be warned though, the MAX11300 spontaneously destroys itself when powered from +-12V. It works for a while, then suddenly blows up with a short circuit to GND. Together wil help from another Eurorack manufacturer we were able to prove this behaviour to the MAXIM support team but they were unable to help us. 
A solution seems to be to restrict the operating voltages to -5 and +12V.

GR8Instruments:
@TheSlowGrowth, thanks for the info.
So far I have not had an issue with the device as you have noted above.

If it helps any here is how I have mine hooked up to power:
AVDD_1 and AVVD_2: from A_+5V  
AVSSIO_1 through AVSSIO_6: from -2V 
AVDDIO_1 through AVDDIO_2: from +12V

The pins are set to 0 to 10V output in my configuration. 

I based my voltage choices off of [page 10](https://datasheets.maximintegrated.com/en/ds/MAX11300.pdf)
I targeted: DAC Range and ADC range to be 0 to 10V putting the power at: VAVDDIO = +12V, VAVSSIO = -2V

In looking at that chart the only thing that would be close to +12V and -12V would be setting the chip  VAVDDIO = +12V and VAVSSIO = -10V and that would give you DAC range 0V to +10V and the ADC range of -10 to 0V.  

So what you are saying about powering it from +12/-12V can make some sense to me given that chart.  But, to be fair, I am only on rev 2 of my device... so I may still run into this issue.  

Thank you for the pointer to the code and the heads up that I may have an issue.

GR8Instruments:
@TheSlowGrowth,

Thank you for the info above.  Currently I am using -2 and +12 so that I can zero to +10 for the DAC output.  So far that has been working well.

I do have a question for you as I ran into this issue when linking:
...
cannot convert 'daisy::MAX11300Driver<daisy::MAX11300MultiSlaveSpiTransport, 1>::PinConfig::<unnamed union>' to 'daisy::MAX11300Types::DacVoltageRange'
...
For now, given I am using only one output range of zero to 10 I hard coded the DAC zero to 10 output and it does compile (I have not run it yet).

In my BSP package I have created the following:
- Setup of my BSP Header I have the following:

```
    daisy::MAX11300<1>              max11300;
    daisy::MAX11300<1>::Config      max11300_config;
    daisy::MAX11300Types::DmaBuffer maxDMA;
```
Sample of what is in the hardware initialize:
...
In my hw setup I have the following: 

    // Setup the max for DAC outs for CV/GATES A and B
    max11300_config.Defaults();
    max11300.Init(max11300_config, &maxDMA);

    // Gate A
    max11300.ConfigurePinAsAnalogWrite(0, MAX11300Types::PIN_8, MAX11300Types::DacVoltageRange::ZERO_TO_10);
    max11300.WriteAnalogPinVolts(0, daisy::MAX11300Types::PIN_8, 0.0f);

    // the rest of my setup
    // ...

    // at the end of that I call
    max11300.Start();
... 

Inside of the audio loop (where I had my older update) I put in the following for now: 
...
            hw.max11300.WriteAnalogPinVolts(0, Seq.Tracks[track].DACPin, Seq.Tracks[track].osc.Process());
...

After it builds I received the error above:
...
...
cannot convert 'daisy::MAX11300Driver<daisy::MAX11300MultiSlaveSpiTransport, 1>::PinConfig::<unnamed union>' to 'daisy::MAX11300Types::DacVoltageRange'
...

Any thoughts for feedback here on possible ways to solve this would be appreciated. 

Thanks,
Brett

PR #473, opened 2022-01-10
Vl53l0x device wrapper

beserge:
Port the device, this time bringing over the files as in and only changing those things strictly necessary


stephenhensley:
closing because its old, and we haven't really tested it -- there also hasn't been much demand for this -- I'll leave the branch intact

PR #477, opened 2022-01-26
Hall Sensor device

beserge:
Simpler Hall Sensor class based on Switch / GPIO rather than the STM32 hall sensor peripheral.


stephenhensley:
closing because its old, and we haven't really tested it -- there also hasn't been much demand for this -- I'll leave the branch intact

PR #483, opened 2022-02-16
Add ChannelModeEvent and AsChannelMode() to midi

### LINKED ISSUE #395: [MIDI] No way to parse event AsChannelMode
MIDI CC numbers 120-127 are considered `ChannelMode` events. It's possible to look at an event's type to determine that it's a ChannelMode , but there's no way to then parse `AsChannelMode` in order to determine which ChannelMode message you received. For that, you need to parse `AsControlChange` and look at the event's `control_number`. For example, you can parse ChannelMode 123 `AsAllNotesOff` but you'd first need to parse is `AsControlChange` to determine that it was in fact 123. Parsing a CC event gives you all the information you need to take appropriate action, so parsing again as something else isn't necessary.

beserge:
fixes #395 

Not sure we want to close the issue yet. I've got some simple code up on feature/midi_distributor which starts the process we were discussing there.

This PR just handles the ChannelMode stuff.

### DIFF
--- a/src/hid/MidiEvent.h
+++ b/src/hid/MidiEvent.h
@@ -57,7 +57,7 @@ enum SystemRealTimeType
 
 enum ChannelModeType
 {
-    AllSoundOff = 0,     /**< & */
+    AllSoundOff,     /**< & */
     ResetAllControllers, /**< & */
     LocalControl,        /**< & */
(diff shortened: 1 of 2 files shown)


stephenhensley:
Looks good to me!

PR #484, opened 2022-02-17
Changes for issues with compiling with optimisation

AndrewCapon:
There are issues with the HAL usb files: stm32h7xx_ll_usb.h and stm32h7xx_ll_usb.cpp where there can be timing issues due to code being optimised out. 

See: https://community.st.com/s/question/0D50X0000BWpRO7SQN/stm32h7-usb-driver-fails-highest-optimization-o3

The solution from that link was to set any io memory locations to _IO and make the delay count variables volatile. I have done this for these files.


stephenhensley:
@AndrewCapon thanks for the report, and quick fix!!

I think it looks OK to me.

---

To clarify a bit, when using the default linker, and running everything from the internal flash the USB works properly even with `-O3` set?

Does the failure occur with either USB port, or just one (built-in, or external).

If that's the case, it's interesting that it only becomes an issue when the firmware is executing from SRAM. Are you using the bootloader in `core/`, and one of the other included linker scrips, or do you have some sort of custom set up running? 

I ask because we have seen one other existing issue with that bootloader that we've yet to identify/resolve where one of the external USB pins doesn't function properly after exiting the bootloader.

AndrewCapon:
p.s. Has anyone been working on the USB Audio, I worked on some code a few years ago for an stm32f7 that had Composite audio and midi. I may be able to dig that out...

PR #487, opened 2022-02-21
Uart periph improvements

beserge:
- Added missing UART4 periph pins
- Added defaults to Config struct


stephenhensley:
Approved!

PR #488, opened 2022-02-22
SPI Improvements

beserge:
- Added SPI2 pins needed for patch_sm
- Added default values in Config constructor


stephenhensley:
Looks good!

PR #498, opened 2022-03-03
Bootloader v5

CorvusPrudens:
This PR adds the latest version of the Daisy bootloader. It also updates the core makefile to address the correct USB PID, which has been changed to `0xA360`.


stephenhensley:
We'll probably want to work a bit more on the guide over time (change app types to a table, screenshots, etc.) but for now I think it's passable.

Everything else seems okay.

stephenhensley:
Just as some final cleanup I:

* did a bit of formatting in the guide
* updated the linker for the SRAM bootloader to not fill into the reserved SRAM used in the bootloader
* updated to latest version of the bootloader that's being tested

Also I think we could do an update where SRAM can be programmed via openocd as long as we remove the reset by changing:

```
	$(OCD) -s $(OCD_DIR) $(OCDFLAGS) \
		-c "program ./build/$(TARGET).elf verify reset exit"
```

to

```
	$(OCD) -s $(OCD_DIR) $(OCDFLAGS) \
		-c "program ./build/$(TARGET).elf verify exit"
```

Or something similar, but I haven't tested this yet. So we can try it later.

PR #500, opened 2022-03-17
Fix debounce rate limit issue

beserge:
When debounce rate is being limited, you would get multiple `true` when you checked the edge between debounces.
I added an `updated_` flag to only send an edge or inc once per debounce


stephenhensley:
looks good, and seems to work good!!

PR #507, opened 2022-04-01
Added virtual function to detect user interaction.

ifranco:
I need a way to detect any interaction with the UI system. The use case was to be able to have a page close itself after a timeout if there's no user input.
I believe the event needs to propagate to all the pages, so it uses a loop triggered right after ProcessEvent(e) inside the Ui::Process function. Maybe it should be inside the if that checks for an invalid event type?

### DIFF
--- a/src/ui/UI.cpp
+++ b/src/ui/UI.cpp
@@ -60,11 +60,10 @@ void UI::Process()
             lastEventTime_        = currentTimeInMs;
             UiEventQueue::Event e = eventQueue_->GetAndRemoveNextEvent();
             if(e.type != UiEventQueue::Event::EventType::invalid)
-                ProcessEvent(e);
-            
-            for(int32_t i = pages_.GetNumElements() - 1; i >= 0; i--)
             {
-                pages_[i]->OnUserInteraction();
+                ProcessEvent(e);
+                for(int32_t i = pages_.GetNumElements() - 1; i >= 0; i--)
+                    pages_[i]->OnUserInteraction();
             }
         }
     }
--- a/src/ui/UI.cpp
+++ b/src/ui/UI.cpp
@@ -61,6 +61,11 @@ void UI::Process()
             UiEventQueue::Event e = eventQueue_->GetAndRemoveNextEvent();
             if(e.type != UiEventQueue::Event::EventType::invalid)
                 ProcessEvent(e);
+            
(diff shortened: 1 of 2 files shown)


src/ui/UI.h
```diff
@@ -92,6 +92,9 @@ class UiPage
     /** Returns true if the page is currently active on a UI - it may not be visible, though. */
     bool IsActive() { return parent_ != nullptr; }
 
+    /** Returns true if an event is being processed. */
+    virtual bool OnUserInteraction() { return true; }
```
TheSlowGrowth, line 96:
```suggestion
    /** Called on any user input event, after the respective callback has completed. OnUserInteraction will be invoked for all pages in the page stack and can be used to track general user activity. */
    virtual void OnUserInteraction() {}
```

src/ui/UI.cpp
```diff
@@ -61,6 +61,11 @@ void UI::Process()
             UiEventQueue::Event e = eventQueue_->GetAndRemoveNextEvent();
             if(e.type != UiEventQueue::Event::EventType::invalid)
                 ProcessEvent(e);
+            
+            for(int32_t i = pages_.GetNumElements() - 1; i >= 0; i--)
```
TheSlowGrowth, line 65: This loop needs to sit inside the `if(e.type != UiEventQueue::Event::EventType::invalid)` statement. An Invalid event returned from the queue means: there was no event to pull - so consequently there also was no user activity.

TheSlowGrowth (changes requested):

src/ui/UI.h
ifranco: Updated.

src/ui/UI.cpp
ifranco: Updated.

TheSlowGrowth (approved):

stephenhensley:
It's looking good!

Definitely a useful feature!

I've done some similar things with some external mechanics added into the UI stuff on some fairly recent projects. So this will certainly come in handy!

Thanks for the contribution!
I'll do a final check and get it merged a bit later!

ifranco:
> It's looking good!
> 
> Definitely a useful feature!
> 
> I've done some similar things with some external mechanics added into the UI stuff on some fairly recent projects. So this will certainly come in handy!
> 
> Thanks for the contribution! I'll do a final check and get it merged a bit later!

Thanks! My pleasure.

PR #511, opened 2022-04-18
USB Host callbacks

stephenhensley:
Adds callback mechanism to the USB Host class.

This allows for running the `USBHostHandle::Process()` function in the main while loop as a background task, and writing specific callbacks for dealing with connection, disconnection, etc.

At the moment we're only handling one class (MSC), so the callbacks are pretty straight forward, but once we have add a multiple classes (audio, HID, etc.) we'd have to add some breaking change to make those easy to parse, handle separately. 

So it might be worth just setting up some of the class checking data now, and updating the `USBClassActive` callback to be class-specific now.
Alternatively, we could pass in the USBHost data itself and let the user parse it.

As an aside, using the USBHostHandle class, and FatFS results in binaries too large to debug without using the custom bootloader.

### DIFF
--- a/Middlewares/ST/STM32_USB_Host_Library/Class/MSC/Src/usbh_msc.c
+++ b/Middlewares/ST/STM32_USB_Host_Library/Class/MSC/Src/usbh_msc.c
@@ -253,7 +253,7 @@ static USBH_StatusTypeDef USBH_MSC_InterfaceDeInit(USBH_HandleTypeDef *phost)
 
   if (phost->pActiveClass->pData)
   {
-    USBH_free(phost->pActiveClass->pData);
+    // USBH_free(phost->pActiveClass->pData);
     phost->pActiveClass->pData = 0U;
   }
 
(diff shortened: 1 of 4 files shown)


stephenhensley:
I want to do a separate guide in the docs for working with USB host, and this may warrant a few upcoming changes as well.

In the projects I'm working on so far I'm mimicking the internal, unused `AppliState_TypeDef` for my own application in the user callbacks, and then using those to drive a state machine to mount the filesystem, etc.

So probably a bit of stuff to rework.

Also, the `Result` codes aren't entirely helpful because the blocking functions always return `USBD_OK` if stuff is functioning, and I haven't found a great way to detect when all activity (especially diskio) is actually complete as it seems there is a bit of lag between when something like `f_close` returns and the subsequent `USBHost::Process()` finish negotiating the completion of activity.

But for now it works, and allows a user application to support replugging USB while running which wasn't possible before with the existing API.

Issue #506, opened 2022-03-30
The SSD130x4WireTransport is lacking full configuration


stephenhensley:
The I2C transport is solid, but the SPI transport only exposes the 2 auxiliary GPIO (Reset, and Data/Command) for configuration.

The `SpiHandle::Config` object should be added to the internal `Config` struct similar to how it's setup in the I2C transport.

The current hardcoded values can be moved to the `Defaults` function in the `Config` struct. Since anything that has been successfully using this code would still want to have those defaults.

At some point both the I2C and SPI transport `Defaults` functions should be converted to constructors as well.

stephenhensley:
Fixed in #512

PR #512, opened 2022-04-25
Add the SpiHandle:Config struct to SSD130x4WireTransport:Config

recursinging:
This is a fix for issue #506. It seems a number of people have been running into this, and it was an oversight on my part, so I fixed it.  

@stephenhensley - I added constructors to the `Config` structs that invoke `Defaults()` as you mention in the issue description.  This should be fully backward compatible, but I don't have time to test against an SPI transport OLED.  Perhaps you could quickly verify this against a Daisy Patch?  Otherwise it's little more than a little code shuffling, and should not introduce any breaking changes.

Cheers,
Sam


stephenhensley:
@recursinging Thanks, Sam!

This looks great, I saw some of the forum activity you're talking about, and was hoping we'd be able to carve out some time this month to get this in there.

Thanks for making it happen!

Looks good to me. We'll do a quick check that everything still works with the existing Daisy Patch code, and get it merged.

PR #515, opened 2022-04-28
Midi Junk Status byte bugfix

beserge:
Previously, if we received a status byte when we were expecting data, we would throw it away and wait for the next message.
However, in this case we should go back to the beginning of the parser and actually handle that status byte.
I wrote a test that checks for this behavior. It's worth noting we did fail the test before the change!

### DIFF
--- a/src/hid/midi.h
+++ b/src/hid/midi.h
@@ -253,6 +253,7 @@ class MidiHandler
                 {
                     // invalid message go back to start ;p
                     pstate_ = ParserEmpty;
+                    Parse(byte);
                 }
                 break;
             case ParserHasData0:
@@ -269,6 +270,12 @@ class MidiHandler
 
                     // At this point the message is valid, and we can add this MidiEvent to the queue
                     event_q_.Write(incoming_message_);
+                }
+                else
+                {
+                    // invalid message go back to start ;p
+                    pstate_ = ParserEmpty;
+                    Parse(byte);
                 }
                 // Regardless, of whether the data was valid or not we go back to empty
                 // because either the message is queued for handling or its not.
--- a/src/hid/midi.h
+++ b/src/hid/midi.h
@@ -156,7 +156,7 @@ class MidiHandler
     void Parse(uint8_t byte)
     {
         // reset parser when status byte is received
-        if((byte & kStatusByteMask) && pstate_ != ParserSysEx) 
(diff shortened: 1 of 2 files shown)


tests/Midi_gtest.cpp
```diff
@@ -638,5 +638,40 @@ TEST_F(MidiTest, badData)
...
+{
+    uint8_t noteon            = 0x90;
+    uint8_t running_status[6] = {0x40, 100, 0x40, 0x01, 0x40, 100};
+    uint8_t status_bytes[3]   = {0x90, 0x91, 0xB3};
+
+    Parse(&noteon, 1);
+    for(int i = 0; i < 3; i++)
+    {
+        MidiEvent   event   = ParseAndPop(&running_status[i * 2], 2);
+        NoteOnEvent noEvent = event.AsNoteOn();
```
TheSlowGrowth, line 655: `noEvent` sounds very much like "no event". Would `noteOnEvent` or `onEvent` make it easier to read?

src/hid/midi.h
```diff
@@ -253,6 +253,7 @@ class MidiHandler
                 {
                     // invalid message go back to start ;p
                     pstate_ = ParserEmpty;
+                    Parse(byte);
```
TheSlowGrowth, line 256: This recursive approach has a risk of unwanted side effects:
If the second status byte in a row is received, we
1. reset the state and enter the `Parse()` function again
2. properly parse the status byte, because now we're in the right section of the `switch`
3. return from `Parse()` and get back here 
4. process the remaining parts of the initial call to `Parse()`

After step 3 we would already be done. Step 4 is only there because of the recursion and it could easily break the parser if someone adds additional code to the end of the `Parse()` function.

Instead, you could also add this to the beginning of the parser:
```cpp
void Parse(uint8_t byte)
{
    // reset parser when status byte is received
    if((byte & kStatusByteMask)) 
    {
        pstate_ = ParserEmpty;
    }
    switch(pstate_)
    {
        // ...
    }
}
```

TheSlowGrowth (changes requested):

src/hid/midi.h
beserge: Good catch! One interesting thing to note is you have to ignore the supposed "status byte" when you're receiving sysex, since it might just be part of the sysex data being sent.

Caught that with the unit test. They're the best!

stephenhensley (approved):

PR #518, opened 2022-04-28
User callbacks to timer peripheral

stephenhensley:
This was useful in the case where there were low priority "background" tasks that need to happen at a regular interval (UI updating, Eventqueue handling, etc.), but actual background tasks that only function with blocking I/O for now (i.e. QSPI writes, and USB Host background process).

Without having the UI, etc. in an IRQ they would stutter/stop/glitch during the QSPI writes/USB processing.

In the rare case where there's a decent amount of processing happening in these callbacks, the Systick was getting missed. So I bumped the NVIC priority of systick to have one position below it. 

I'd like to add a priority setting to these callbacks so the user can set it up at the desired level, but for now these callbacks have the lowest priority in the whole system.

### DIFF
--- a/src/per/tim.cpp
+++ b/src/per/tim.cpp
@@ -164,9 +164,9 @@ TimerHandle::Result TimerHandle::Impl::Stop()
 
 TimerHandle::Result TimerHandle::Impl::SetPeriod(uint32_t ticks)
 {
-    config_.period = ticks;
+    config_.period                = ticks;
     tim_hal_handle_.Instance->ARR = ticks;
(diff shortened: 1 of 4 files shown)


stephenhensley:
TODO:

* [ ] Add IRQ enable, etc. for remaining peripherals
* [ ] Add configurable IRQ Priority (0-15) to match NVIC config.
* [ ] add `enable_irq` bool to config to avoid unnecessary IRQs if not desired
  * this seems a better fit than having it be based on whether the callback is set or not since the user can change that at will.
* [ ] Add doxygen documentation because it's all missing currently

PR #523, opened 2022-05-18
TimerHandle callback fixes/improvements

stephenhensley:
This accomplishes a few related things:

* Implements the user-callback for the rest of the supported, generic TIM instances (TIM2, TIM3, TIM4)
* Fixes bug where DAC DMA/Callback usage would crash due to null `Impl::Config` object in internal hardware init
* Adds example of using the new `TimerHandle::PeriodElapsedCallback`

### DIFF
--- a/examples/TIM_SingleCallback/Makefile
+++ b/examples/TIM_SingleCallback/Makefile
@@ -1,0 +1,12 @@
+# Project Name
+TARGET = TIM_SingleCallback
+
+# Sources
+CPP_SOURCES = TIM_SingleCallback.cpp
+
+# Library Locations
+LIBDAISY_DIR = ../..
+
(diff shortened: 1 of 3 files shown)


stephenhensley:
Tested:

* Daisy Patch SM Envelope example (uses TIM6 based DAC callback)
* new TIM5 based callback example
* Tested modified example using TIM4 peripheral
* Tested modified example using TIM3 peripheral
* Tested modified example using TIM2 peripheral*

Re. TIM2

Opened new issue (#524) to describe weirdness with attempting DeInit prior to Init.

For clarity, when using TIM2 (which is internally used by the System class), I had to Init the TIM2 peripheral, then DeInit before initializing again to have a clean state to start a callback from.

This particular use is outside the scope of this PR.

PR #525, opened 2022-05-18
Add bugfix for single data byte running status, add unit test

### LINKED ISSUE #521: Running status not handled for single byte commands
https://github.com/electro-smith/libDaisy/blob/04dd33614ab67fde6deae3ec0bc60a0ad128e3c3/src/hid/midi.h#L217

**Issue:**
Currently here we do not handle running status for single-byte commands (Program Change, MTC Quarter Frame,Song Select, and really, more importantly, _Channel Pressure_). This appears to cause dropped channel pressure messages resulting in a "stuck" pressure status.

**Method of reproduction:**
Using a Yamaha Motif XF, play and hold a chord, and quickly press more firmly and release while still holding the chord to quickly engage/release the channel pressure repeatedly. The synthesizer appears to occasionally send running status messages for channel pressure. If at the very least you track the 'latest' channel pressure message, this should always reset to 0 once all of the keys are released, however currently it may not, as single byte running status messages are being dropped.

**Suggested fix:**
I appear to have fixed this by updating this code to:
```
incoming_message_.type    = running_status_;
incoming_message_.data[0] = byte & kDataByteMask;
//check for single byte running status, really this only applies to channel pressure though
if(running_status_ == ChannelPressure
   || running_status_ == ProgramChange
   || incoming_message_.sc_type == MTCQuarterFrame
   || incoming_message_.sc_type == SongSelect)
{
    //Send the single byte update 
    pstate_ = ParserEmpty;
    event_q_.Write(incoming_message_);                        
}
else
{
    pstate_                   = ParserHasData0; //we need to get the 2nd byte yet.
}

```

beserge:
fixes #521

### DIFF
--- a/src/hid/midi.h
+++ b/src/hid/midi.h
@@ -222,7 +222,21 @@ class MidiHandler
                     // Handle as running status
                     incoming_message_.type    = running_status_;
                     incoming_message_.data[0] = byte & kDataByteMask;
-                    pstate_                   = ParserHasData0;
+                    //check for single byte running status, really this only applies to channel pressure though
+                    if(running_status_ == ChannelPressure
+                       || running_status_ == ProgramChange
+                       || incoming_message_.sc_type == MTCQuarterFrame
+                       || incoming_message_.sc_type == SongSelect)
+                    {
+                        //Send the single byte update
+                        pstate_ = ParserEmpty;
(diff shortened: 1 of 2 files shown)


stephenhensley (approved):
Looks good here, and passes the new tests.

PR #530, opened 2022-05-31
Feature: audio output compensation

stephenhensley:
Minor feature update for the `AudioHandle`.

This adds the ability to set a compensation value for the output signal to match the analog input level.

This can be useful in hardware where the input and output levels differ due to the attached circuits.

It is the equivalent of doing a multiply on each output sample at the end of the callback.

This differs from the "postgain" which is an adjustment to the "1.0" point to allow for hardware with extra headroom to have its nominal analog level match a normalized -1 to 1 range in software.

---

Tested this on hardware using 24-bit SAI bit-depth, non-interleaving callback

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Unreleased
 
-<!-- Add New Changes Entries Here -->
+### Features
+
+* audio: added `output_compensation` value to config struct to allow for post-scaling of uneven audio passthru levels.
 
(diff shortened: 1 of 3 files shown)


TheSlowGrowth:
Just saw this now. Could we maybe precompute the `postgain * compensation` to avoid introducing an additional multiplication step for each sample?

stephenhensley:
opened a new PR #531

PR #531, opened 2022-06-01
Precompute audio output adjust

stephenhensley:
Just a slight optimization to avoid the added per-sample multiplication added with the new `AudioHandle::Config` parameter.


TheSlowGrowth (approved):
Looks good to me :-)

PR #536, opened 2022-07-19
Bugfix/fix max11300 analog output

TheSlowGrowth:
There was a compiler error in `Max11300Driver::WriteAnalogPinVolts()`.

This PR fixes that error and modifies the unit tests to use that function at least once so that an error like this can't slip though again.


stephenhensley (approved):

stephenhensley:
Somehow took half the year to notice this was done, and waiting. Sorry about that!

Looks good.

Issue #534, opened 2022-06-30
SRAM Bootloader Breaks SW1 and SW2 on Daisy Field


byronellis:
Hi, switching to the SRAM-based boot loader (BOOT_SRAM) appears to break SW_1 and SW_2 on the Daisy Field. Specifically it seems like RisingEdge() is always true. Not sure if this is intentional or if there is a workaround, but the same code will work fine if you load the same code without the custom daisy boot loader.

stephenhensley:
@byronellis thanks for the bug report, we'll look into this!

byronellis:
No problem! If you need a minimal reproduction I can make one but it's pretty much "switches don't seem to work" so it's almost trivial 🙂

dromer:
Ok can confirm that with this branch of libDaisy and the bootloader the switches work as expected.

Really hope this gets merged/released soon, but at least I know the cause of the issue and that it will get resolved at some point.

dromer:
I suppose this ticket can now be counted as resolved. Or should it need a release tag for libDaisy first?

dromer:
This works now on bootloader v6 and beyond :)

PR #548, opened 2022-11-07
Bootloader v6.0

CorvusPrudens:
This adds the Daisy bootloader v6 to libDaisy, which fixes any stuck-pin issues such as #534. This also adds a parameter to the `System::ResetToBootloader` method which allows returning to the Daisy bootloader (both with the timeout, and without).


ohmtech-rdi (commented):
It seems to me that the changes to `src/daisy_seed.cpp` should be reflected as well in `src/daisy_patch_sm.cpp`.

(At least I see that when using the bootloader v6 and a program on QSPI, my program doesn't run, but didn't check in the debugger yet)

When I update my code template to reflect the changes in `src/daisy_seed.cpp`, it works at least on my end:
https://github.com/ohmtech-rdi/eurorack-blocks/commit/7b6d5cd9c05fcca1b90355e126decd8a90ba58c2

ohmtech-rdi:
@CorvusPrudens When do you think this will be merged? Currently using the previous bootloader makes it not really usable when having a program on QSPI.

If this will take some time, do you know if there is a workaround we could apply on our end please?

Thanks

stephenhensley:
Hi @ohmtech-rdi 

Can you describe the issue you're experiencing with running a program from QSPI on the v5 bootoader (and if that still persists on this branch)?
Can you also clarify if you're using the `BOOT_SRAM` app type, or the `BOOT_QSPI` app type when you run into issues?

The more recent commits here should have everything working for the PatchSM at this point.

I'd just like to make sure that we're aware of, and resolving any existing issues as we go through testing this PR, and staging it for release.

ohmtech-rdi:
Hi @stephenhensley 

The problem is described a bit more in detail [here](https://github.com/ohmtech-rdi/eurorack-blocks/issues/469#issuecomment-1367339257)

But in essence, when using `BOOT_QSPI`, at least the `A8` pin, that we use as a GPIO out in eurorack-blocks kivu12 for the address of the multiplexer of our pots, couldn't be changed for some reason.

Switching to the v6 bootloader solves everything for us (if there was other pins than `A8`, we didn't check), users are happy as far as I know, so that's why we are so eager to see this branch merged.

OmsInSerial:
Hi,
When do you think this will be merged? I need it for my module to work properly.
Regards

dromer:
@stephenhensley how is it going with this, will it happen before christmas? :santa:

stephenhensley:
@dromer yes! This should be merged by the end of the week.

We're wrapping up some testing with a few new configurations of the bootloader now, and then those new bin files, and a little update to the core/Makefile are all that need to happen.

Worth mentioning that the new configurations are available for testing in a [test repo](https://github.com/stephenhensley/UsbBootloaderTest) I made. Pending any issues we haven't found yet, those bin files will be the ones coming in.
The intusb, 2000ms timeout configuration is the equivalent of the current version of the bootloader bin staged on this branch.

beserge:
Tested the example under examples/BootLoaderBlink with the default + all four custom BOOT_BIN settings.
Works great! :shipit:

CorvusPrudens:
I also tested all configurations of the example successfully!

stephenhensley:
Everyone who's tested this, including myself haven't run into any issues.

Thanks everyone for the contributions, testing, and feedback in general as we worked on this update!

Merging now! 🚀😄

PR #550, opened 2022-11-24
Fix MCP23x17.h reading of multiple registers at once

benishor:
Modified size of data being read when attempting to read both A and B registers at once.


brbrr (approved):
LGTM

stephenhensley:
@benishor Thanks for the PR!
@brbrr Thanks for jumping on the review.

It's looking good here as well!

PR #551, opened 2022-12-01
SSD130x4WireSoftSpiTransport

eh2k:
At the moment the hardware SPI2 on the Submodule does not seem to work properly. After some trial and error, using an OLED display via software spi is a viable alternative for me.  The advantage is also that any gpio pins can be used. In contrast to spi_handle::BlockinTransmit no "ScopedIrqBlocker" is necessary.

### DIFF
--- a/examples/OLED_SoftSPI/Makefile
+++ b/examples/OLED_SoftSPI/Makefile
@@ -1,0 +1,14 @@
+# Project Name
+TARGET = OledSoftSPI
+
+# Sources
+CPP_SOURCES = OledSoftSPI.cpp
+
+# Library Locations
+LIBDAISY_DIR = ../../
+
+#APP_TYPE = BOOT_SRAM
+
+# Core location, and generic Makefile.
+SYSTEM_FILES_DIR = $(LIBDAISY_DIR)/core
+include $(SYSTEM_FILES_DIR)/Makefile
--- a/examples/OLED_SoftSPI/OledSoftSPI.cpp
+++ b/examples/OLED_SoftSPI/OledSoftSPI.cpp
@@ -26,26 +26,30 @@ MyDisplay display;
 
 int main(void)
 {
-	// Initialize the hardware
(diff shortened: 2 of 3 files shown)


stephenhensley:
@eh2k thanks for the PR! This looks great.

Awesome that you included in an example as well!

The style-stuff can be fixed pretty easily with the fix style script if you have python and clang-format installed. 
Otherwise, there's not a whole lot to fix by hand if you don't.
Let me know if you need any help with that.

What issues did you run into getting SPI2 to work on the Patch SM? Was a particular pin not functioning properly?

GregBurns:
FYI, SPI2 has been working fine for me on the PatchSM to drive an OLED
display. However it doesn't play well with USB. I have been unable to get
the logger and display to work together. I posted a small test program on
the Daisy forum.

eh2k:
@stephenhensley 

My current issue/findings:

I have the SM rev 3, and with the current libDaisy and BOOT_SDRAM the display stays black when I control it with the hardware SPI transport. I only got this to work with SpiHandle::Config::BaudPrescaler::PS_128 and if I used the ScopedIrqBlocker around the display.Update(). As far as I can remember, the ScopedIrqBlocker was only necessary when executing hw.StartAudio(). I also examined the whole thing with a LogicAnlyser and found that at the end the SPI sequence is somethig strange with the clock (at the end there is an unnecessary clock change). 

image


I also noticed that the SPI ports D9, D8 are read by the ADC. I have removed these testwise from the ADC initialization (without success).
After some googling I found the stm32h7 errata, where it says that there are SPI problems at high clock rates. After further investigation I found out that the whole clock configuration in the System::Init() is skipped in the BOOT_SDRAM case, or probably done in the bootloader (closed source).

stephenhensley:
The soft transport looks good. So I'm going to merge it.

I've split out the notes about your issues with the hardware SPI to new issue #553

PR #556, opened 2022-12-26
default config fix for MidiUartTransport

drum-computer:
Hey this is my very first pull request ever :) 
I've been adding midi in port to daisy and somehow it wasn't working, so I've noticed in the midi.h that default `config `of `MidiUartTransport` wasn't referencing correct pin numbers. When I changed those numbers according to the daisy datasheet my midi input started working.

### DIFF
--- a/src/hid/midi.h
+++ b/src/hid/midi.h
@@ -34,8 +34,8 @@ class MidiUartTransport
         Config()
         {
             periph = UartHandler::Config::Peripheral::USART_1;
-            rx     = {DSY_GPIOB, 7};
-            tx     = {DSY_GPIOB, 6};
+            rx     = Pin(GPIOPort::PORTB, 15);
+            tx     = Pin(GPIOPort::PORTB, 14);
         }
     };
 


stephenhensley:
Hi @drum-computer thanks for the PR!
Excited to see that you're interested in contributing back to the Daisy libraries!

To clarify, the `MIDIUartTransport::Config` has the default pins set up for the pins exposed on the Daisy Seed and Patch SM pin-banks. 
These can be overridden when creating the config struct by manually setting:

```cpp
MidiUartTransport:::Config midi_cfg;
midi_cfg.rx = Pin(kMyRxPin);
midi_cfg.tx = Pin(kMyTxPin);
// and then initialize
my_midi_object.Init(midi_cfg);
```

The pins that you configured to the defaults are valid pins for the USART1 peripheral, but all of the example hardware platforms with MIDI use the pins already set as defaults.

If you're using the Patch SM, what I've described above would work for your projects, but does not require an update to the library itself.
You can also use PB8/PB9 via UART4 on the PatchSM.

Feel free to let me know if you have any questions!

PR #558, opened 2023-01-04
Correct Patch SM gate out pin assignments

CorvusPrudens:
This PR corrects some gate out pin assignment confusion added by (#417) as noted by [apbianco](https://forum.electro-smith.com/u/apbianco) and [tele_player](https://forum.electro-smith.com/u/tele_player) for the Patch SM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # libDaisy Changelog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


stephenhensley:
Confirmed proper "gate_out_1" and "gate_out_2" initialization as well as B5, B6 assignment.

PR #561, opened 2023-01-09
Update Patch SM Pin use to be consistent with modern API

stephenhensley:
updated patchsm pin objects to use newer Pin object instead of old dsy_gpio_pin

This'll also update the gate_out objects to be the new style GPIO.

Being a "breaking change" this will update the major version of libDaisy when merged, but migration should be simple:

1. For any existing objects that still take `dsy_gpio_pin` instead of `daisy::Pin`, there is a built-in operator that will work
2. For actual GPIO, the new pins can still initialize old GPIO, so a new object `dsy_gpio gateout` could still be initialized with the new Pin object.

And any existing gate output work on the Patch SM would change from:

```cpp
dsy_gpio_write(&hw.gate_out_1,true);
dsy_gpio_write(&hw.gate_out_1,false);
dsy_gpio_toggle(&hw.gatte_out1);
```
 to:

```cpp
hw.gate_out_1.Write(true);
hw.gate_out_1.Write(false);
hw.gate_out_1.Toggle();
```

### DIFF
--- a/src/daisy_patch_sm.cpp
+++ b/src/daisy_patch_sm.cpp
@@ -6,51 +6,51 @@ namespace daisy
 namespace patch_sm
 {
     /** Const definitions */
-    static constexpr dsy_gpio_pin DUMMYPIN        = {DSY_GPIOX, 0};
-    static constexpr dsy_gpio_pin PIN_ADC_CTRL_1  = {DSY_GPIOA, 3};
-    static constexpr dsy_gpio_pin PIN_ADC_CTRL_2  = {DSY_GPIOA, 6};
-    static constexpr dsy_gpio_pin PIN_ADC_CTRL_3  = {DSY_GPIOA, 2};
-    static constexpr dsy_gpio_pin PIN_ADC_CTRL_4  = {DSY_GPIOA, 7};
-    static constexpr dsy_gpio_pin PIN_ADC_CTRL_5  = {DSY_GPIOB, 1};
-    static constexpr dsy_gpio_pin PIN_ADC_CTRL_6  = {DSY_GPIOC, 4};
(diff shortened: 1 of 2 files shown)


stephenhensley:
To hold backwards compatibility, and avoid the breaking change we could add new `GPIO` objects and leave the old ones in place.

However, they'd have to be named differently, which I believe would lead to more confusion than anything else.

For now I'm leaving the gate objects as old `dsy_gpio` and we can choose to make that change now or later.z

src/daisy_patch_sm.cpp
```diff
@@ -6,51 +6,51 @@ namespace daisy
...
-            {DSY_GPIOA, 0},  /**< A3  - UART1 Tx */
-            DUMMYPIN,        /**< A4  - GND */
-            DUMMYPIN,        /**< A5  - +12V Power Input */
-            DUMMYPIN,        /**< A6  - +5V Power Output */
-            DUMMYPIN,        /**< A7  - GND */
-            {DSY_GPIOB, 14}, /**< A8  - USB DM */
-            {DSY_GPIOB, 15}, /**< A9  - USB DP */
-            DUMMYPIN,        /**< A10 - +3V3 Power Output */
+            DUMMYPIN,       /**< A1  - -12V Power Input */
+            Pin(PORTA, 10), /**< A2  - UART1 Rx */
```
stephenhensley, line 28: I think we bypassed our full testing on the PatchSM/Patch.Init earlier in the process (when we were talking about it duplicating the efforts in the existing PR).

Just noticed that this pin got converted incorrectly; it should be `Pin(PORTA, 1)`

We'll go through a quick test on the patch to see if there's anything else that we missed, but this looks like the only issue.

stephenhensley:
This is now a duplicate of #581. Since everything here is taken care of there, and tested. I'll close this.

PR #563, opened 2023-02-06
docs: use explicit grouping; omit comments from output

scolsen:
While browsing the rendered API docs I noticed that section comments were being passed to the output (e.g. // start hid). This commit fixes that by doing two things:

- wrapping all section comments with a conditional tag on an IGNORE variable. This will prevent doxygen from including them in output unless the IGNORE variable is set when rendering.
- Use the explicit @ingroup command to define group nesting, instead of @{ }@ blocks. This isn't strictly necessary but I feel it is a bit cleaner and makes it harder to mess things up when it comes to tabbing/spacing. In addition, it allows the section comments to be included more clearly amidst the actual section definitions.

### DIFF
--- a/doc/extra/index.h
+++ b/doc/extra/index.h
@@ -1,64 +1,100 @@
 #pragma once
 namespace daisy {
-// Define the groups to be used throughout all of libdaisy for 
+// Define the groups to be used throughout all of libdaisy for
 // automated documentation
 /** @defgroup libdaisy LIBDAISY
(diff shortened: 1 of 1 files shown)


stephenhensley:
@scolsen thanks for the PR! This looks great, definitely a lot easier to read than it was.

I'm trying to get a bunch of things ready for a release later this week. So I'll bring this in with the rest of the updates as well.

scolsen:
Sounds good!

PR #565, opened 2023-02-07
chore: remove redundant compile-time def

scolsen:
Removes a redundant compile-time definition in the cmake toolchain file. STM32H750xx was set twice, once in a bundle of other definitions, and a second time by itself.


stephenhensley:
@scolsen sorry for the delay in getting to this, but looks sensible to me!
thanks for the PR!

Issue #570, opened 2023-02-14
Bug: memory in address space mapped to QSPI flash does not update when app is running from Daisy Bootloader


ndonald2:
### Expected behavior

When running an application from the Daisy bootloader (e.g. `APP_TYPE = BOOT_SRAM`), using `QSPIHandle` to write to flash mapped to memory should update the contents of the mapped memory space.

### Actual behavior

When running an application from the Daisy bootloader (e.g. `APP_TYPE = BOOT_SRAM`), using `QSPIHandle` to write to flash mapped to memory does not update the contents of the mapped memory space while the app is running. but the mapped memory contents will be correct on the next device reboot.

### Repro steps

1. Build and flash Daisy Bootloader
2. Ensure app is built using `APP_TYPE = BOOT_SRAM`
3. Write some code to write any known data to flash using `QSPIHandle` to an address offset of zero (should be safe according to docs, since it's well before `0x90040000`)
4. Use `QSPIHandle::GetData()` to read back the data at address offset zero from mapped memory
5. Note that the contents of the mapped memory does not match the data you just wrote
6. Reset/reboot the Daisy device
7. Repeat step 4
8. Note that the data _does_ match the previously written value now

Repeat the above with the default internal flash linkage, no daisy bootloader and note that the mapped memory contents _are_ updated after a flash write.

### Notes

* I was testing this on a Seed
* I will happily provide more details / isolate a minimal example project as requested.
* I noticed this trying to use code derived from `PersistentStorage` (though this also manifests when using that class directly), because every call to `StoreSettingsIfChanged()` was resulting in a write because the comparison always returns that there are changes, even when there shouldn't be - until the next device reboot, that is.

stephenhensley:
Thanks for the detailed report.
This is very strange behavior indeed.

I wonder if there's some caching-issue happening when running from SRAM that doesn't happen when running from flash.
Could be a unique interaction of the D-Cache while reading from SRAM.
The flash and AXI do have different device memory area attributes (as per table 7 in the reference manual) that could be the cause.

QSPI is write-through
AXI SRAM is write-back write-allocate
Flash is write-through

I can do a bit more digging on this this week, but in the meantime, I'd suggest trying the following if you have time:

1. Does disabling the DCache `SCB_DisableDCache()` for the entire program result in the same behavior.
2. If that resolves it then you could try running `dsy_dma_invalidate_cache_for_buffer` on the chunk on the data at the address you're using from QSPI.

[...]

PR #572, opened 2023-02-14
Fix QSPI memory cache invalidation for PersistentStorage

ndonald2:
### Summary

Band-aid fix for #570 

### Details

If the program is running outside internal flash, we need to invalidate the DCache for the memory mapped to QSPI before comparing settings data, otherwise it will always be "dirty" until the next device reboot.

See linked issue #570 for more details.


ndonald2:
Noticed the change broke unit tests so I did the simplest thing I could think of and skipped the cache invalidation code when building for unit tests 🤷  LMK if you think there's a better solution. I wasn't sure how best to stub/mock the dependencies on `System` and the cache invalidate function.

stephenhensley:
@ndonald2 as ugly as ifndefs can be, I think this a suitable fix for now. It's very clear why it would be getting skipped.

Thanks so much for the fix!

thealienthing:
Hey all! I'm a student doing a project on daisy seed and I've been having this very issue but I have the latest changes in libdaisy with this PR included. I'm not certain as to what part of memory my firmware is being stored since I'm still very new to stm32 development, but if it helps understand, I'm pretty much using my seed in the most default way I know of by pressing the boot and reset buttons and flashing the firmware with make program-dfu. So I believe my program is in fact running within internal flash. 

Either way, I'm having the same behavior of not seeing changes to my flash data between qspi erase/write actions without performing a reboot. Any ideas where I should look to fix this? I appreciate any help I can get.

stephenhensley:
Hi @thealienthing 

So the fix here applied only when **not** running from internal flash. It's possible that there's a similar/related issue when running from internal flash.

To see if the same solution resolves the problem, you can do either of the following:


1. Comment out lines 129-130 of the `src/util/PersistentStorage.h` and recompile both libDaisy, and your project.

or 

2. manually invalidate the cache before attempting to read back the data from the reference provided by `PersistentStorage<>::GetSettings()`:

```
SCB_InvalidateDCache();
```

Let me know if either of those work, and we can make a patch to get it fixed.

Thanks!

thealienthing:
Excellent! I appreciate your quick response. I'll give this a try tonight and let you know if that solves the problem.

PR #574, opened 2023-03-10
Add driver for Adafruit Dotstar LEDs (SK9822)

ndonald2:
### Summary

Adds device support for SK9822 addressable RGB LEDs (aka Adafruit Dotstar) with SPI Transport

### Details

This is a device support "driver" class for the SK9822 RGB addressable LED. The code patterns are modeled after other device classes in libDaisy, implemented as a template class with data transport broken out separately. Currently there is only an SPI transport implementation but in theory one could add transport support for other peripherals.

The device class is meant to simultaneously manage the color states of 1 or more serially connected LEDs or "pixels" (currently up to a hard-coded maximum of 64, but this could probably be made a template parameter). The nature of the device is such that all LEDs must be updated at once, since the serial data must be clocked through the entire chain, so it's simplest to just keep all the data in memory until ready to display.

The usage should be fairly self explanatory based on the doc comments but an example is provided below.

### Possible Future Improvements

The SPI transport currently uses blocking transmits rather than DMA. In practice this is _probably_ OK for the typically small number of LEDs in synth designs and the fast clock speed in use (default baud prescaler of 4 which results in a clock rate of 6.25MHz seems to work fine for me). But it might be nice to have the option to do DMA based SPI writes instead.

Additionally, the `Color` class in libDaisy could be improved to handle many common tasks for working with Color data (in the context of addressable LEDs or otherwise):

- HSV color space support, including methods to modify HSV on an existing instance
- Gamma correction
- Static methods to get named colors rather than `Init`
  - In general would be nice if the class behaved more like a POD value type, with constructors etc (breaks libDaisy convention somewhat but makes up for it with convenience).

### Example Usage

For patch submodule, which uses SPI2 instead of the default SPI1

```c++
#include "daisy_patch_sm.h"

using namespace daisy;
using namespace daisy::patch_sm;

const uint8_t kNumLeds = 4;

static DaisyPatchSM hw;
static DotStarSpi dotstars;

int main(void)
{
	hw.Init();
	hw.StartLog();

	DotStarSpi::Config dotstar_cfg;
	dotstar_cfg.Defaults();
	dotstar_cfg.transport_config.periph   = SpiHandle::Config::Peripheral::SPI_2;
	dotstar_cfg.transport_config.clk_pin  = Pin(PORTD, 3);
	dotstar_cfg.transport_config.data_pin = Pin(PORTC, 3);
	dotstar_cfg.color_order               = DotStarSpi::Config::ColorOrder::RBG; // If colors seem off, change this
	dotstar_cfg.num_pixels 	              = kNumLeds;

	dotstars.Init(dotstar_cfg);
	dotstars.SetAllGlobalBrightness(1);

	float b = 0.0f;
	uint32_t val = 0;
	while(1)
	{
                // cube of ch brightness for gamma correction
		val = b * b * b * 255;
		dotstars.SetPixelColor(0, val, 0, 0);  // red
		dotstars.SetPixelColor(1, 0, val, 0); // green
		dotstars.SetPixelColor(2, 0, 0, val); // blue
		dotstars.SetPixelColor(3, val, 0, val); // pink
		dotstars.Show();
		b += 0.01f;
		if (b >= 1.0f) b = 0.0f;
		System::Delay(33);
	}
}
```

### DIFF
--- a/src/util/color.cpp
+++ b/src/util/color.cpp
@@ -9,12 +9,12 @@ using namespace daisy;
 
 namespace daisy
 {
-    template <typename T>
-    T clamp(T in, T low, T high)
-    {
-        return (in < low) ? low : (high < in) ? high : in;
-    }
+template <typename T>
+T clamp(T in, T low, T high)
+{
+    return (in < low) ? low : (high < in) ? high : in;
 }
+} // namespace daisy
 
 const float Color::standard_colors[Color::LAST][3] = {
     {1.0f, 0.0f, 0.0f},   // RED
--- a/src/util/color.cpp
+++ b/src/util/color.cpp
@@ -1,9 +1,20 @@
 #include "util/color.h"
+#include <algorithm>
+
 using namespace daisy;
 
 #define RED 0
 #define GREEN 1
 #define BLUE 2
+
+namespace daisy
+{
+    template <typename T>
+    T clamp(T in, T low, T high)
+    {
+        return (in < low) ? low : (high < in) ? high : in;
+    }
+}
 
 const float Color::standard_colors[Color::LAST][3] = {
     {1.0f, 0.0f, 0.0f},   // RED
@@ -25,7 +36,7 @@ void Color::Init(PresetColor c)
 
 void Color::Init(float red, float green, float blue)
 {
-    red_   = red;
-    green_ = green;
-    blue_  = blue;
+    red_   = clamp(red, 0.0f, 1.0f);
+    green_ = clamp(red, 0.0f, 1.0f);
+    blue_  = clamp(blue, 0.0f, 1.0f);
 }
--- a/src/daisy.h
+++ b/src/daisy.h
@@ -50,6 +50,7 @@
 #include "dev/mcp23x17.h"
 #include "dev/max11300.h"
 #include "dev/tlv493d.h"
+#include "dev/dotstar.h"
 #include "dev/neopixel.h"
 #include "dev/neotrellis.h"
 #include "dev/icm20948.h"
--- a/src/dev/dotstar.h
+++ b/src/dev/dotstar.h
@@ -58,9 +58,10 @@ class DotStarSpiTransport
 
     void Write(uint8_t *data, size_t size)
     {
-      if(spi_.BlockingTransmit(data, size) != SpiHandle::Result::OK) {
-        printf("uh oh");
-      }
+        if(spi_.BlockingTransmit(data, size) != SpiHandle::Result::OK)
+        {
+            // TODO: Error handling
+        }
     };
 
   private:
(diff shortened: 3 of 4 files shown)


src/util/color.cpp
```diff
@@ -25,7 +36,7 @@ void Color::Init(PresetColor c)
 
 void Color::Init(float red, float green, float blue)
 {
-    red_   = red;
-    green_ = green;
-    blue_  = blue;
```
ndonald2, line 30: I also added clamping for the color values here - there isn't a generic clamp function available in libDaisy AFAICT so I yanked the templated version from `leddriver.h`

stephenhensley: weird. I could have sworn we at least had an assembly-optimized floating point clamp function in here, but I guess that's in the DaisySP side of things. 

Fine to leave it here for now, though!

stephenhensley:
@ndonald2 As always, thanks for the fantastic contribution!

This looks good to me! I don't have anything to test locally, but assuming it drives dotstars okay, I think this is probably ready to merge once you're satisfied. 

Re. color improvements, I agree with what you suggest.
Semi related, I actually think I might have a branch lying around somewhere that has some nice QoL stuff like basic constructors, and some operator overloads for adding, scaling, and blending. 
I think for utilities (esp. non-audio-rate) convenience can definitely outweigh convention. So I'm definitely open to that.

I'll see if I can find that sometime this week, and I'll make a PR.

src/util/color.cpp
ndonald2: I will take another look just to be sure. There is one in DaisySP (in dsp.h) but I couldn't find anything equivalent in libDaisy.

stephenhensley:
@ndonald2 re. having another set of eyes: Up to you. If you've been using it, then I'm sure its good enough to start with, and we can always apply a patch if someone catches something.

And super cool! I'll take a look around Friday or this weekend to see if I can find that code and post an initial PR.
I don't think I went and did the `constexpr`ing, but that would be a nice addition for sure.

ndonald2:
@stephenhensley cool cool, I'm satisfied so feel free to merge this anytime. I haven't exercised every part of the interface but all the basic stuff is working great (set constant-current brightness, clear/fill, set individual pixel colors, draw...)

If there's anything I'm slightly torn about it's exposing the full 5-bit range of constant current brightness in the interface, since really only the first 5 levels or so are practical for most use cases due to power consumption and thermal concerns. I guess the warning comment is enough? But I don't want anyone to fry their circuits by trying to ramp it up really high.

stephenhensley:
@ndonald2 Cool, well I think its ready then! I'll hit merge.

> If there's anything I'm slightly torn about it's exposing the full 5-bit range of constant current brightness in the interface, since really only the first 5 levels or so are practical for most use cases due to power consumption and thermal concerns. I guess the warning comment is enough? But I don't want anyone to fry their circuits by trying to ramp it up really high.

This is a good point, but as long as the default is sensible, I'd suspect a user would see the warning at the same time that they see the function for updating it. So I think it's fine.

Thanks again for the PR!

PR #579, opened 2023-04-27
Add ConversionSpeed configuration to ADC class

stephenhensley:
This PR is based on the repo at v3.0.0, but only changes the ADC initialization.

This adds the ability to change the conversion time of each channel of the configured ADC.

The default value is set to what was previously the fixed value of 8.5 cycles.

This number of cycles is a little misleading because factors like bit depth, etc. also affect the number of cycles it takes to convert in addition to this value. 
The reference manual covers this in a bit more depth, but the settings are named to match the STM32 HAL, and their docs.

There are some vestigial commented-out values of trying fixed values before adding the configuration that should now be removed prior to merging.

### DIFF
--- a/src/per/adc.cpp
+++ b/src/per/adc.cpp
@@ -162,7 +162,7 @@ static int get_num_mux_pins_required(int num_mux_ch)
         return 0;
 }
 static void
-write_mux_value(uint8_t chn, uint8_t idx, uint8_t num_mux_pins_to_write);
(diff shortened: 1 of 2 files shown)


stephenhensley:
Pretty minimal update, but it's been tested by myself, and a few other users.
No changes required by users unless they want to override the default (i.e. previously fixed setting) with the new config option.

Issue #538, opened 2022-08-02
libDaisy doesn't compile with recent arm toolchain versions on linux


mfc:
see full details: https://github.com/electro-smith/DaisyWiki/issues/33

stephenhensley:
@mfc thanks for opening the issue!

We're aware of the compatibility issue, and it's the case with the most recent version of the compiler on all OSes.
Fixing it requires updating the entire STM32 HAL to a more recent, compatible version since libDaisy uses that extensively.

So while we do plan on getting this fixed, we don't have a timeline on exactly when it'll get done. 

We'll keep any further progress noted here!

beserge:
Fixed via #582

PR #578, opened 2023-04-18
added arm_clip to cmsis library

danielkinahan:
I needed this in my project so I'll add it here.

PR #582, opened 2023-05-22
Update internal CMSIS and HAL

stellar-aria:
This upgrades the built-in CMSIS library to v5.9.0, the CMSIS H7 Device to v1.10.3, and the STM32 HAL to v1.11.1.

This addresses issue #538.

This change has been tested with DaisyExamples and a Daisy Patch, and tested against the included test suite.

### Additions:
The new HAL modules have been added to `src/sys/stm32h7xx_hal_conf.h`, and have been enabled to match the others already present. These include modules such as the Digital Temperature Sensor, a Filter Math Accelerator (aka Fixed-point Multiplier and Accumulator or FMAC), an Octo-SPI controller (OSPI), a Digital Filter for Sigma-Delta modulation, and of course, the CORDIC co-processor block, which can accelerate trigonometric functions like sine, cosine, arctan, and square root for q1.31 and q1.15 fixed point signed numbers.

Makefiles have been updated to reflect the source code structure changes, including `Makefile`, `core/Makefile`, `CMakeLists.txt`, and the `libdaisy.vcxproj[.filters]` files.

### Major Breaking Changes:
`GPIO::Mode::OUTPUT_OD` has been renamed to `GPIO::Mode::OPEN_DRAIN` due to a name conflict collision with a macro now defined in stm32h7xx_hal_gpio.h (the primary HAL file for gpio). This is instead of a kludge like `#undef`ing the macro, as it may crop up elsewhere depending on include order.

Documentation has been updated accordingly.

### Minor Breaking Changes:
Compiled code size has increased by up to 7%. This means that the `patch/Nimbus` example in DaisyExamples will not compile any longer and may need to be reworked to be stored somewhere other than FLASH.

### Minor Fixes:
A very minor (single-character) bugfix was done in `CpuLoadMeter_gtest.cpp` so that a number would no longer overflow when bitshifting left, and so that I could actually compile and run the tests.

The relevant external repos:
https://github.com/ARM-software/CMSIS_5
https://github.com/STMicroelectronics/cmsis_device_h7
https://github.com/STMicroelectronics/stm32h7xx_hal_driver


Drivers/STM32H7xx_HAL_Driver/Inc/stm32h7xx_hal_gpio.h
```diff
@@ -286,7 +282,25 @@ void HAL_GPIO_EXTI_Callback(uint16_t GPIO_Pin);
...
+#define GPIO_MODE_Pos                           0u
+#define GPIO_MODE                               (0x3uL << GPIO_MODE_Pos)
+#define MODE_INPUT                              (0x0uL << GPIO_MODE_Pos)
+#define MODE_OUTPUT                             (0x1uL << GPIO_MODE_Pos)
+#define MODE_AF                                 (0x2uL << GPIO_MODE_Pos)
+#define MODE_ANALOG                             (0x3uL << GPIO_MODE_Pos)
+#define OUTPUT_TYPE_Pos                         4u
+#define OUTPUT_TYPE                             (0x1uL << OUTPUT_TYPE_Pos)
+#define OUTPUT_PP                               (0x0uL << OUTPUT_TYPE_Pos)
+#define OUTPUT_OD                               (0x1uL << OUTPUT_TYPE_Pos)
```
stellar-aria, line 294: This is the location of the offending macro that causes name-collision

stellar-aria:
Hi! Thanks for the feedback, I've added a comment to the line number of the file which causes the name-collision.

As you said, readability is also a concern, if just because it's better to be explicit for users, however if this wasn't a critical compilation failure I'd be fine just leaving it as it was.

Yeah, I don't really think there's a quick easy fix for the HAL's size creep, especially if the Daisy platform adopts other CPUs in the future (for say, a Seed 2 or 3) and doesn't want to take on the task of writing custom platform-specific adapters between the higher-level libDaisy classes and the processor-specific LL stuff. It's all tradeoffs, and the HAL is overall just more convenient.

Drivers/STM32H7xx_HAL_Driver/Inc/stm32h7xx_hal_gpio.h
stellar-aria: Since macros are basically just global find-and-replace text replacement, the preprocessor is inserting it into the enum regardless of the code (and scoping) around it :/

stephenhensley: Ah, yup! You're totally right. I forgot that macros just do a brute-force text replace everywhere. This is totally fine.
Thanks for the added detail.

stephenhensley:
Thanks for the update. 
I think we'll be able to have someone on the team start doing some testing on this next week.

We'll keep you posted if there's anything that needs changing.

stellar-aria:
After giving it a lot of thought, the DSP subfolder of the CMSIS driver should probably also be swapped out for a submodule and pinned to a release tag now that it's been spun off into its own project. It'll also keep the code turnover in this PR down once the commits are squashed and merged.

stellar-aria:
This will also make #578 unnecessary, as the DSP library will be up-to-date and now includes arm_clip_f32 (at https://github.com/ARM-software/CMSIS-DSP/blob/3a04f817a4380c00ff51f9bcdc6c4e4bb7a18b80/Include/dsp/basic_math_functions.h#L821)

stellar-aria:
Yeah, I actually re-incorporated those noted edits into the updated version of the HAL driver files. If there's problems with the SDMMC DMA stuff, it'll be because I didn't see any notes and it got undone 😅. I'll do a diff of the driver pre-and-post and check what's going on in a few minutes.

In terms of switching to submodules, we can probably at least go ahead with the CMSIS driver as well then, since the preliminary commit here was just based on the source at the latest release tag.

stellar-aria:
Okay, after some thinking at lunch, the easiest thing to do would probably be to swap to using a submodule, but retain those singular patched source files in-tree until the changes are merged upstream, at which point the build managers can just be pointed back to the originals.

This is what I'm going to work on for now, unless something changes?

Alternatively, the daisy project could fork the HAL, introduce the required bugfixes, and then submit a PR or just keep the separate fork and then have that be the submodule.

stellar-aria:
It looks like the CI isn't set up to pull submodules, and that's why it's failing now.

```
Run actions/checkout@v2
  with:
    repository: electro-smith/libDaisy
    token: ***
    ssh-strict: true
    persist-credentials: true
    clean: true
    fetch-depth: [1](https://github.com/electro-smith/libDaisy/actions/runs/5248697620/jobs/9481319644?pr=582#step:2:1)
    lfs: false
    submodules: false
    set-safe-directory: true```

stellar-aria:
Another potential source of regressions is the Middleware drivers now being out of date, but unless they do cause problems, upgrading them is probably best left for a future PR. 

So I'm just noting some things here for the future:

The USB Device Middleware can probably be upgraded (and swapped to a submodule) without any issues. 

However there's some notes on the Host library that dynamic allocation was disabled for one of the internal Handle types, and then later notes that USBH_Free was also disabled, and the main barrier to using the Middlewares as-designed (with malloc and free) was that the heap is in DTCMRAM.  As far as I can tell the only memory configuration with the heap located there is the SRAM one, but I also don't think I understand the architecture well enough yet to understand why it was placed there initially.

stellar-aria:
Hey just wondering where everything's at, I figure testing this is a real pain. I see there's some merge conflicts from upstream, should I go ahead and resolve those or will that impact the build that's being used?

stephenhensley:
Hi @stellar-aria 

You can resolve the conflicts, that shouldn't cause any issues with testing.
Testing has been going okay so far. 

I'm hoping to get dig into it quite a bit deeper this week, and do a few updates on some older, but more complex programs that use USB, SDMMC, etc. and do some side-by-side comparison of some of the non-trivial changes. 
That should validate the new versions work consistently (or better) than the older versions.

If nothing breaks/doesn't stand out, then I think we might be able to merge this this week 😄

stellar-aria:
Since I Introduced this PR over six months ago, I've learned a lot more about the state of embedded development toolchains and build systems, primarily from my work managing such for the [Deluge Community Firmware](https://github.com/SynthstromAudible/DelugeFirmware) project. In that time, my overall outlook w.r.t. the management of dependencies (such as the CMSIS ones here) and build systems in the Daisy project has changed, and it boils down to a few key points:

1) Maintaining two build systems is a hassle and can lead to problems. Make and CMake ultimately accomplish the same thing, but CMake offers far more flexibility in terms of project organization and integration into other projects as a library. From what I understand the CMake system is newer, and is potentially a migration target, but (much like with the GPIO/Pin situation) currently in an awkward middle state.

2) Submodules are not the greatest way to do dependency management. CMake now has a _very_ nice built-in dependency download and integration system called [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) that is well suited for problems like this: downloading weird out-of-source build components and integrating them without polluting the version control system. I've used this in a very similar situation for downloading and integrating AVR Dx devicepacks into a project's AVR-GCC toolchain, for example.

In the long run migrating the current build system to more idiomatic CMake is probably a good path forward, and will reduce a lot of issues with other kinds of integrations such as alternate toolchains (I've had a Clang-support PR waiting in the wings for a while now). However in the short term I think submodules are an acceptable stopgap solution for the dependency version problem, even if it requires an additional VCS step in the form of `git submodule update`.

I also kind of just want to see this PR merged so I can work on others, haha 😅

stephenhensley:
Hi @stellar-aria 

Happy New Year!
Sorry for the delay in getting to this, the last few months of the year became very busy for us.

We are doing some internal testing on this, and that should progress fairly quickly. We are aiming to have either approval, or in the case of finding any issues, feedback by the end of next week!

Regarding your comments on CMake, I've converted that into and issue (#606) so that we could discuss it separately since it is a worthwhile discussion, but a bit outside the scope of this PR.

stellar-aria:
Just checking in on this!

stephenhensley:
@stellar-aria alright! 

@beserge and I have both gone through quite a bit of testing at this point.
Overall, things are working great. However, there were a few things that Ben and I both experienced that are worth taking a bit more of a look at:

### USB Device

I am unable to use the internal USB port (I haven't tested the external yet) with the updates. 
It does trigger the first few IRQ handlers, but then seems to get stuck in a busy state. I only checked this with the CDC class/Logger class setup. I know Ben tested with MIDI and ran into a similar issue of it not showing up on his PC.

I have a feeling this could be an incompatibility in our class interface with at least one of the changes that have happened over the years. 
I did see that there is a new "TransmitCplt" callback added to the interface operations for the API, although ST's patch notes mention that a subsequent version made this optional/backwards compatible with previous versions. 

Did you run into this on your end with either of the device classes?

### USB Host

Similar to the above, I setup some test code that worked on the current master branch that no longer works correctly on the new branch.
However, I am still getting some USB Host behavior (connect/disconnect awareness), but it does end up failing in FatFS with disk errors.

Again, I would guess this is an incompatibility with the latest changes (interestingly it looks like the USB Host middleware has been left the same. So it might be on the HAL side that's causing this particular issue.

### SDRAM/FMC

So I haven't personally gotten any abnormal behavior yet, but Ben mentioned having an example not start up. Afterwards it worked okay.

I haven't done any strict profiling to see if there's been any unexpected change in performance yet either.
Just wanted to make a note about that here in case you've run into any other similar events, or have done any time-based profiling on SDRAM performance.

---

I'm going to dig into the USB stuff a little bit today/tomorrow to see if I can find the particular issue(s) remaining there. If you have any insight from testing you've done so far that would be welcome!

Aside from getting USB up and running, I'm thinking the rest of this is ready to go.

stellar-aria:
I haven't played much with the USB stuff yet, tbqh. Part of the incompatibility might just come from using the older middleware libraries with the newer HAL, as I noted earlier:
> Another potential source of regressions is the Middleware drivers now being out of date, but unless they do cause problems, upgrading them is probably best left for a future PR.
> 
> So I'm just noting some things here for the future:
> 
> The USB Device Middleware can probably be upgraded (and swapped to a submodule) without any issues.
>
> However there's some notes on the Host library that dynamic allocation was disabled for one of the internal Handle types, and then later notes that USBH_Free was also disabled, and the main barrier to using the Middlewares as-designed (with malloc and free) was that the heap is in DTCMRAM. As far as I can tell the only memory configuration with the heap located there is the SRAM one, but I also don't think I understand the architecture well enough yet to understand why it was placed there initially.

So it might just be worth it to upgrade the device library and see if that fixes the device problems. If it does, we might have to upgrade/merge the host library manually. It doesn't look like the heap is stored in the DTCMRAM anymore, rather in SRAM for the QSPI and Flash loaders, and the D2 RAM for the SRAM loader, so does that note need to be edited/fixed?

I haven't done any strict profiling for the SDRAM, if you need something like that I do have some profiling tools. Was this code execution from the SDRAM or just data storage there?

stephenhensley:
Okay!

I have a feeling you're correct, I just wanted to confirm if you had either experienced the same thing, or had a different experience before we went super far down the rabbit hole.
We'll test around with updating the device libraries, and go from there.

As far as I can tell, the USB device/host issues are the only things left to resolve, and then I think this is good to go.

---

I don't think there's much to do wrt to the SDRAM, the performance seemed pretty similar to me. The only red flag was the report of an app not starting, but that could have been something unrelated as it has not been reproduced since.

I didn't do any quantitative profiling, but I have a program that just barely underruns when copying the input around the SDRAM a ton of times, and then to the output, and it was audibly the same when running it between both the master branch, and this PR. 
Other tests, and code using the SDRAM for data worked as usual as well.

stephenhensley:
@stellar-aria Just wanted to keep you up to date as we progress. 

We've made some decent headway in identifying the issues with USB, and are hoping to have a patch for getting USB device stuff working before the end of the week. 
If that's the case, we should have a decent idea of how to get USB host working again as well, and I'm hoping to have that knocked out in the first half of next week. So if that timeline works out, we can potentially be doing final testing, and be ready to merge this by the end of next week.

GregBurns:
FYI I have been working on MIDI host USB. It works on first boot but
unplug/plug cycles are unstable. I am continuing to debug this issue. I can
point you at my git repo if you are interested.

beserge:
@GregBurns Awesome! Please do. I'm going to look into this myself a bit now.

I've just pushed my patches to the USB Device Middleware required to use both USB Midi and CDC.
As part of that I updated the ref for the HAL Driver.

GregBurns:
Ben,
     This is my fork

https://github.com/GregBurns/libDaisy


Greg

beserge:
Ok, so I was finally able to successfully test USB Host and get it working. It was a combination of issues on my end.

1. My example had to be compiled for the bootloader due to size constraints.
2. The Daisy Web Programmer is (I think) still flashing an old version of the bootloader. I made sure to flash an updated version.
3. This branch doesn't include some crucial bootloader stuff if you checkout directly to it. Once I merged with master it worked flawlessly.

Inexplicably the app still was able to be debugged, and the issues didn't present as if they were bootloader related, not really sure why yet.

Couple of things I want to do:
- Confirm which version of the bootloader is being flashed by the web programmer.
- Merge my usb host example into libDaisy

That being said, this PR is finally tested, and ready to be merged! Thanks for your hard work, glad we're finally able to give it the signoff! 🚀

beserge:
Ready for merge! Thanks once again for your _stellar_ contribution!

PR #583, opened 2023-06-13
UART DMA Rework and UART MIDI improvements/refactoring

ndonald2:
### Summary

Extends and supersedes @stephenhensley's work from #557:

* [Breaking Change] UART peripheral to add DMA "listen" mode with idle interrupts to supersede older FIFO approach
* Factor out `MidiParser` to a dedicated class separate from (but used by) `MidiHandler`
    * For backwards compatibility and maintaining existing tests, `MidiHandler` still has a publicly avialable `Parse(uint8_t byte)` method.
* Change `MidiUartTransport` to use new UART DMA mode
* Change `MidiUartTransport` to accept a "parse callback" function and context pointer when starting Rx
    * Callback is called every time the DMA Rx callback fires with new received bytes.
    * Bytes are immediately parsed into FIFO owned by `MidiHandler` rather than waiting for application - as evidenced from testing, this _drastically_ improves practical performance of UART MIDI (e.g. dropped messages)
    * Minor changes to USB transport implementation to conform to the same signature, for `MidiHandler` transport template parameter. USB transport also uses parse callback in a similar way.
* Add option to provide DMA buffer for UART MIDI byte rx in the transport config struct, with sensible defaults
    * Enables this new approach to work correctly for Daisy Bootloader app types
    * (Theoretically) enables applications with multiple MIDI UART inputs on different peripherals, using separate `MidiUartHandler` instances, provided application code provides separate DMA rx buffers for each.
* Minor automatic formatting changes (trim whitespace, add trailing newline to files, etc)

The only breaking changes should be in `UartHandler`. Most existing applications using UART MIDI should not require changes.

### Testing Notes

I tested this using the following setup:

* Seed running a 24-voice MIDI polysynth application example with internal end-of-chain reverb.
* Repeating 8-voice 16th note chords sent to seed over UART MIDI at 120 BPM, with real-time MIDI clock in the stream as well
    * From Bitwig Studio + USB Focusrite interface translating USB MIDI to UART/DIN
    * From Keystep Pro via UART/DIN MIDI output
* Stress tested with randomly automated CC streams in addition to the above
    * At some point the MIDI bus gets overloaded when doing this from the Bitwig setup

This setup was tested across app types (internal flash + daisy bootloader SRAM/QSPI) using default libDaisy linker scripts as well as via USB MIDI directly for regression testing purposes.


stephenhensley:
I just pushed a new unit test that confirms one of the remaining bugs with MIDI.

It tests the parser by running a full note-on message, followed by several clock messages (system real time), and then the two data bytes for a 0-velocity note on (note off) message.
This seems to be how the Elektron boxes function (specifically the Octatrack), and currently causes note-off messages to get missed.

@beserge is going to add a fix to the newly relocated midi parser on this branch.
Once that fix is applied, the unit test should pass, and I _think_ this PR will be ready to merge.

stephenhensley:
Alright!

@ndonald2 @beserge thanks for getting this where it is!

I think at this point the things we set to do are accomplished, and we got a few other things fixed as well.

In addition to the UART rework, and moving parsing into the actual interrupt process we caught, and resolved a bug with the parser that had to do with SystemRealtime messages improperly invalidating the running status byte.

This PR has been tested with several configurations. 
All three built-in linker/application types (flash, SRAM, and QSPI) were tested.
The rxbuffer was tested at many sizes (down to two bytes, which is not recommended as it will result in a lot of stuck notes/missing data), and memory sections (uncached d2 ram, AXI SRAM, SDRAM, and heap allocated via malloc).

Performance is much better than the current master branch.

This is approved! 🚀

PR #584, opened 2023-06-14
Global GPIO/Pin Migration

stellar-aria:
This continues the work started in #581, updating references to `dsy_gpio_pin` and `dsy_gpio` to instead use the new C++ style GPIO and Pin classes.

As per the discussion with @stephenhensley, this is a breaking change for board header files (their GPIO variables are now updated), but does not include any of the commits removing the now-obsolete, deprecated code. As such, the API as a whole remains largely unaltered. The commits dealing with code removal will be submitted in a future PR once the documentation work discussed has been completed and preparations begin for a new major revision.
 
### Minor breaking API changes:
DaisySeed, DaisyPatch, daisy::patch_sm, and DaisyField now use `GPIO` instead of `dsy_gpio`.


stephenhensley:
Hi @stellar-aria 

Thanks for another stellar (pardon the pun) PR. 
Quickly looking through everything I think everything looks good.
So exciting to see some of the oldest bits of the library catching up with the times.

There are a few more in-progress PRs that will get wrapped up/merged within the next week (MIDI fixes, the ADC speed feature, etc.) as part of a new minor release.
Once that's out, I think this would be a good first step toward the larger-scope cleanup including the HAL update in #582, etc. that we'd been discussing. 

I'll keep you updated if we find anything here that needs changing, but unless anything got missed, or there are any typos, I think this might just be good to go.

stellar-aria:
Just merging in the upstream changes to make sure everything integrates cleanly. Had to do some conflict resolution to get everything to mesh smoothly.

stellar-aria:
This now matches up with the latest master and should be a clean merge.

stephenhensley:
@stellar-aria thanks for syncing this up! We're starting this week on review/testing on this one (should go a _lot_ faster than the HAL stuff).

We'll keep you posted if anything comes up.

beserge:
Made some changes:
- Used the new Pin class exclusively in the DaisySeed class
- Change the GetPin functions in DaisySeed and DaisyPatchSM to return  daisy::Pin
- Remove the old version of the hal_map utils
- Remove the old dsy_gpio_pin based daisy::GateIn::Init()
- Remove the dsy_gpio_pin based functions from gpio.cpp. This version will only support using dsy_gpio_pin as a temporary struct to then be immediately converted to a daisy::Pin
- Make some changes to dsy_gpio_pin as pre deprecation steps
  - Mark dsy_gpio_pin as deprecated, this throws warnings in gcc now if you use it
  - Get rid of some of the helpers for dsy_gpio_pin and the implicit conversion from Pin->dsy_gpio_pin. These were triggering the deprecation warnings, and shouldn't be needed at this point anyways.
  - Add an implicit conversion from dsy_gpio_pin -> daisy::Pin. This required moving dsy_gpio_pin and port into the daisy namespace.

beserge:
Other than the stuff I edited, I went through and tested that everything works as intended. I'm ready to sign off! :shipit:

PR #586, opened 2023-07-08
Allow user to override C_INCLUDES env var in makefile

nullobject:
This change allows a user to modify the `C_INCLUDES` environment variable in their project makefile.

This is similar to the way the `libDaisy` makefile allows users to override `C_SOURCES`, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,7 +289,7 @@ C_DEFS =  \
 # ^ added for easy startup access
 
 
-C_INCLUDES = \
+C_INCLUDES += \
 -I$(MODULE_DIR) \
 -I$(MODULE_DIR)/sys \
 -I$(MODULE_DIR)/usbd \


stephenhensley:
Hi @nullobject thanks for the contribution!

I believe the functionality you're attempting to add is already present in the [`core/Makefile`](https://github.com/nullobject/libDaisy/blob/master/core/Makefile#L127-L129) -- this is the file that is imported from the user's project Makefile.

If you have a particular use-case for adding to the C_INCLUDES of libDaisy itself prior to compilation, we can discuss that further. However, it looks like might have just missed that your requested functionality already works.

nullobject:
Thanks @stephenhensley, you are right. Sorry for the false alarm.

I was looking in `libDaisy/Makefile`, not `libDaisy/core/Makefile`.

PR #589, opened 2023-08-14
Move SAI initialized check so it isn't a no-op

schult:
The current check for an uninitialized `SaiHandle` during DeInit is a no-op: `&config_` will never be null. This change moves the init check out of the `Impl` class so that the `pimpl_` pointer itself can be examined, allowing the check to work as intended. This also fixes a compiler warning about `&config_` never being equal to `nullptr`.


beserge:
Nice catch! Thanks for your contribution, works great!

PR #590, opened 2023-08-23
Add SH1106 OLED driver

nullobject:
This PR adds a driver for the SH1106 OLED display.

The SH1106 is functionally equivalent to the SSD130x, but uses a 132x64 display buffer. This means that we need to override the `Update` method to render into the larger display buffer correctly.


atoav:
I can approve that this works on actual hardware, thanks for the work.

PR #604, opened 2024-01-11
Fix float range for `Random::GetFloat`

CorvusPrudens:
This PR fixes an issue with `Random::GetFloat` following [this discussion on Discord](https://discord.com/channels/1037767234803740694/1039589609505574942/1194379231023607879). I can confirm that before the fix, I got a value above the expected range 4/10 times. Afterwards, it was 0/100.


beserge:
Tested with a quick example, works great!

PR #605, opened 2024-01-16
Fix link in README.md

stellar-aria:
Current link directs to a GoDaddy parking page.


stephenhensley:
oh my goodness! Thanks for the catch!

PR #609, opened 2024-02-09
Add serial read example.

stephenhensley:
This adds a new example showing reading data over a USB serial connection:

This program initializes the daisy, and then waits for connection to a serial monitor.

Once the connection has been established, messages received over serial will be echoed back.
This is done by writing the input to a FIFO of messages as they are received, and then from within the main while() loop, messages are popped from the FIFO, and printed back over serial.


beserge:
Merged with latest master and tested on the Pod, works great!

PR #613, opened 2024-02-21
Change AudioHandle::Config to an aggregate type

stellar-aria:
This changes AudioHandle::Config to be an aggregate type, maintaining the default values but replacing the explicit constructor.

This allows for aggregate initialization in addition to the standard constructor. Usage of the standard (default) constructor is still possible, with the same parameter ordering.

No changes to current code are required.


beserge:
Looks great! Thanks again for the contribution 👍

Issue #606, opened 2024-01-17
[build] - Centralize around a single build system: CMake


stephenhensley:
Since I Introduced this PR over six months ago, I've learned a lot more about the state of embedded development toolchains and build systems, primarily from my work managing such for the [Deluge Community Firmware](https://github.com/SynthstromAudible/DelugeFirmware) project. In that time, my overall outlook w.r.t. the management of dependencies (such as the CMSIS ones here) and build systems in the Daisy project has changed, and it boils down to a few key points:

1) Maintaining two build systems is a hassle and can lead to problems. Make and CMake ultimately accomplish the same thing, but CMake offers far more flexibility in terms of project organization and integration into other projects as a library. From what I understand the CMake system is newer, and is potentially a migration target, but (much like with the GPIO/Pin situation) currently in an awkward middle state.

2) Submodules are not the greatest way to do dependency management. CMake now has a _very_ nice built-in dependency download and integration system called [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) that is well suited for problems like this: downloading weird out-of-source build components and integrating them without polluting the version control system. I've used this in a very similar situation for downloading and integrating AVR Dx devicepacks into a project's AVR-GCC toolchain, for example.

In the long run migrating the current build system to more idiomatic CMake is probably a good path forward, and will reduce a lot of issues with other kinds of integrations such as alternate toolchains (I've had a Clang-support PR waiting in the wings for a while now). However in the short term I think submodules are an acceptable stopgap solution for the dependency version problem, even if it requires an additional VCS step in the form of `git submodule update`.

I also kind of just want to see this PR merged so I can work on others, haha 😅

_Originally posted by @stellar-aria in https://github.com/electro-smith/libDaisy/issues/582#issuecomment-1863483649_

stephenhensley:
I like the idea of centralizing around a single build system a lot.

The primary concern I have is that it may appear more complex to the user for their application build files.

For example, a current minimal Makefile for an application may look like:

```Make
# Project Name
TARGET = Blink

# Sources
CPP_SOURCES = main.cpp

LIBDAISY_DIR = ../libs/libDaisy
DAISYSP_DIR = ../libs/DaisySP

# (optional) Includes FatFS source files within project.
#USE_FATFS = 1

# Core location, and generic Makefile.
SYSTEM_FILES_DIR = $(LIBDAISY_DIR)/core
include $(SYSTEM_FILES_DIR)/Makefile
```

[...]

PR #615, opened 2024-02-22
CMake Improvements

stellar-aria:
Over the past year I've done some work on improving the state of the CMake build system, initially inspired by fixing building on native Windows, and then expanding in scope. 

This is part 1 of the work towards fully migrating to CMake as per #606.

Changes in this PR include:
- Fixing the native build on Windows
- Adding support for LLVM/Clang
- Adopting a similar build configuration process as the current Makefile (see example below)
- Auto-detection of embedded build toolchains
- Build the examples with the composite target `examples` (these are not part of the default `all` target)
- Support for all the GoogleTest unit tests via CTest
- Adding matrix toolchain testing to the Github Actions so we can test for regressions on older toolchains
- Cleaning up the GitHub Actions a bit to use cached tools (for both GCC and LLVM). This makes running the test builds much faster.
- Switch from using `TOOLCHAIN_PREFIX` to the CMake standard `CMAKE_C_COMPILER` for manually selecting a toolchain

Example project:
```cmake
cmake_minimum_required(VERSION 3.26)
cmake_policy(SET CMP0048 NEW)

# Fetch libDaisy
include(FetchContent)
FetchContent_Declare(daisy
  GIT_REPOSITORY https://github.com/stellar-aria/libDaisy
  GIT_TAG cmake-only
)
FetchContent_MakeAvailable(daisy)

# Our project declaration

project(daisytest VERSION 0.0.1)

set(FIRMWARE_NAME daisytest)
set(FIRMWARE_SOURCES
  ${CMAKE_CURRENT_LIST_DIR}/src/test.cpp
)
#set(FIRMWARE_DEBUG_OPT_LEVEL -O0) # (optional) to customize Debug optimization level
#set(FIRMWARE_RELEASE_OPT_LEVEL -O2) # (optional) to customize Release optimization level

# DaisyProject uses FIRMWARE_NAME and FIRMWARE_SOURCES to build a target called ${FIRMWARE_NAME}
include(DaisyProject)

target_include_directories(${FIRMWARE_NAME} PUBLIC include)
```


munshkr:
Thank you so much for working on this! I did a quick test and for some reason found that the binary cmake is creating is much larger than one done using the old Makefile. I haven't digged into it yet but here's the memory map for both:

### Cmake
```Memory region         Used Size  Region Size  %age Used
           FLASH:      104468 B       128 KB     79.70%
         DTCMRAM:          0 GB       128 KB      0.00%
            SRAM:       13644 B       512 KB      2.60%
          RAM_D2:         16 KB       288 KB      5.56%
          RAM_D3:          0 GB        64 KB      0.00%
     BACKUP_SRAM:          12 B         4 KB      0.29%
         ITCMRAM:          0 GB        64 KB      0.00%
           SDRAM:          0 GB        64 MB      0.00%
       QSPIFLASH:          0 GB         8 MB      0.00%
```
### Makefile

```Memory region         Used Size  Region Size  %age Used
           FLASH:       63104 B       128 KB     48.14%
         DTCMRAM:          0 GB       128 KB      0.00%
            SRAM:       12436 B       512 KB      2.37%
          RAM_D2:         16 KB       288 KB      5.56%
          RAM_D3:          0 GB        64 KB      0.00%
     BACKUP_SRAM:          12 B         4 KB      0.29%
         ITCMRAM:          0 GB        64 KB      0.00%
           SDRAM:          0 GB        64 MB      0.00%
       QSPIFLASH:          0 GB         8 MB      0.00%
```

munshkr:
Thanks for the thorough explanation, after posting I figured it was probably an optimization flag :)

stellar-aria:
Oh perfect. Yeah, I probably need to think of a better way to have both sane default optimization levels and propagate user-selected ones. I'll poke at that today.

stellar-aria:
Changes:
- Made a separate "daisy-tests" CMake project for test building/running. This allows IDEs such as VSCode and CLion to be able to build (and run) both the unit tests and the main library side-by-side with different compilers, ideal for TDD/BDD development.
- Fixes the github actions changes arising from this
- No longer relies on "CMAKE_CROSSCOMPILING" flag, which was a bit of hack anyways, since it's usually set by the toolchain file with CMAKE_SYSTEM_NAME when it's "Generic"

asavartsov:
@stellar-aria Thank you very much for your work! I tested it on my project and got an `undefined reference to 'USBH_MIDI_SetReceiveCallback'` linker error. Also there is a warning about changing size of `.text` section.

Here is barebones example that builds fine with the Makefile approach and libDaisy from master but fails witch CMake using latest commit from the PR branch.

```cpp
#include "daisy_seed.h"

using namespace daisy;

DaisySeed hardware;
MidiUsbHandler midiUsb;

int main(void)
{
    hardware.Configure();
    hardware.Init();

    MidiUsbHandler::Config midiUsbConfig;
    midiUsbConfig.transport_config.periph = MidiUsbTransport::Config::INTERNAL;
    midiUsb.Init(midiUsbConfig);

    for(;;)
    {
    }
}
```


```
cmake_minimum_required(VERSION 3.26)
cmake_policy(SET CMP0048 NEW)

# Fetch libDaisy
include(FetchContent)
FetchContent_Declare(daisy
  GIT_REPOSITORY https://github.com/stellar-aria/libDaisy
  GIT_TAG 2826863c573c30c87975b46780df81ec425259b2
)
FetchContent_MakeAvailable(daisy)

# Our project declaration

project(daisytest VERSION 0.0.1)

set(FIRMWARE_NAME daisytest)
set(FIRMWARE_SOURCES
  main.cpp
)

#set(FIRMWARE_DEBUG_OPT_LEVEL -O0) # (optional) to customize Debug optimization level
#set(FIRMWARE_RELEASE_OPT_LEVEL -O2) # (optional) to customize Release optimization level

# DaisyProject uses FIRMWARE_NAME and FIRMWARE_SOURCES to build a target called ${FIRMWARE_NAME}
include(DaisyProject)

target_include_directories(${FIRMWARE_NAME} PUBLIC include)
```


```
$ cmake -B build -DCMAKE_BUILD_TYPE=Release
$ cmake --build build
[  2%] Built target FatFs
[ 61%] Built target STM32H7XX_HAL_DRIVER
[ 64%] Built target STM32_USB_DEVICE_LIBRARY
[ 67%] Built target STM32_USB_HOST_LIBRARY
[ 98%] Built target daisy
[ 98%] Building CXX object CMakeFiles/daisytest.dir/main.cpp.obj
[100%] Linking CXX executable daisytest.elf
/Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: warning: start of section .text changed by 8
/Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/libdaisy.a(usb_midi.cpp.obj): in function `daisy::MidiUsbTransport::Impl::Init(daisy::MidiUsbTransport::Config)':
usb_midi.cpp:(.text._ZN5daisy16MidiUsbTransport4Impl4InitENS0_6ConfigE+0x2e): undefined reference to `USBH_MIDI_SetReceiveCallback'
Memory region         Used Size  Region Size  %age Used
           FLASH:      119496 B       128 KB     91.17%
         DTCMRAM:          0 GB       128 KB      0.00%
            SRAM:       65040 B       512 KB     12.41%
          RAM_D2:       17372 B       288 KB      5.89%
          RAM_D3:          0 GB        64 KB      0.00%
     BACKUP_SRAM:          12 B         4 KB      0.29%
         ITCMRAM:          0 GB        64 KB      0.00%
           SDRAM:          0 GB        64 MB      0.00%
       QSPIFLASH:          0 GB         8 MB      0.00%
collect2: error: ld returned 1 exit status
make[2]: *** [daisytest.elf] Error 1
make[1]: *** [CMakeFiles/daisytest.dir/all] Error 2
make: *** [all] Error 2
```

asavartsov:
Also I was not able to make it work properly with LLVM toolchain on Mac. With DaisyToolchain installed cmake always detected it no matter of `TOOLCHAIN_PREFIX` variable present. I removed DaisyToolchain and installed LLVMEmbeddedToolchainForArm 18.0.0. It partially detected compilers and compiled project but could not produce a bin file, only elf and hex are there and there is no size output.

```
cmake -B build -G Ninja -D CMAKE_BUILD_TYPE=Release -D TOOLCHAIN_PREFIX=/Applications/LLVMEmbeddedToolchainForArm-18.0.0-Darwin-universal
No compiler specified. Falling back to searching PATH...
Could not find arm-none-eabi-gcc. Assuming clang is part of embedded toolchain...
-- The C compiler identification is Clang 17.0.6
-- The CXX compiler identification is Clang 18.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/homebrew/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/LLVMEmbeddedToolchainForArm-18.0.0-Darwin-universal/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done (62.6s)
-- Generating done (0.1s)
-- Build files have been written to: /Users/aleksei/ScratchPad/DaisyExamples/seed/Blink/build
```

```
$ cmake --build build
[a bunch of warnings from libDaisy code]
[204/204] Linking CXX executable daisytest.elf
```

```
$ ls -A1 build
.ninja_deps
.ninja_log
CMakeCache.txt
CMakeFiles
_deps
build.ninja
cmake_install.cmake
daisytest.elf
daisytest.hex
```

stellar-aria:
@asavartsov looks like I missed a file in the merge conflict resolution, whoops. Thanks for finding that! That should fix the undefined error for the linker.

Instead of using `TOOLCHAIN_PREFIX` when you're using the auto-detecting toolchain files, you should instead use the standard `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` command-line settings. This integrates much better with IDEs. The autodetection script will then use that compiler instead of hunting for its own (which as you noticed, tries finding GCC first and then falling back to clang).

asavartsov:
Thanks you! It worked perfectly both for linker error and compiler options. When you mentioned IDE I opened my project in VSCode and it gave me similar command, I should have tried it first 🤦 

About `bin` file generation, it was as simple as `set(DAISY_GENERATE_BIN true)`

I cannot thank you enough again, CMake/LLVM is such a huge improvement for me because I use autogenerated DSP code which compiles forever with gcc and just in seconds with clang.

asavartsov:
I also tested CMake with bootloader targets (sram and qspi) and looks like they won't boot from bootloader without `-DBOOT_APP` which is set by Makefile but not by CMake.

Adding `target_compile_definitions(daisy PRIVATE BOOT_APP)` to my project file fixes that (though it probably should be automatic based on `DAISY_STORAGE` value) but this define also breaks Debug CMake target if using GCC toolchain:

```
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal.c.obj): in function `HAL_Init':
[build] stm32h7xx_hal.c:(.text.HAL_Init+0x6c): undefined reference to `D1CorePrescTable'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: stm32h7xx_hal.c:(.text.HAL_Init+0x70): undefined reference to `SystemD2Clock'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: stm32h7xx_hal.c:(.text.HAL_Init+0x74): undefined reference to `SystemCoreClock'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal.c.obj): in function `HAL_InitTick':
[build] stm32h7xx_hal.c:(.text.HAL_InitTick+0x64): undefined reference to `SystemCoreClock'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal_dma.c.obj): in function `HAL_DMA_IRQHandler':
[build] stm32h7xx_hal_dma.c:(.text.HAL_DMA_IRQHandler+0x1a8): undefined reference to `SystemCoreClock'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal_rcc.c.obj): in function `HAL_RCC_ClockConfig':
[build] stm32h7xx_hal_rcc.c:(.text.HAL_RCC_ClockConfig+0x360): undefined reference to `D1CorePrescTable'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: stm32h7xx_hal_rcc.c:(.text.HAL_RCC_ClockConfig+0x364): undefined reference to `SystemD2Clock'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: stm32h7xx_hal_rcc.c:(.text.HAL_RCC_ClockConfig+0x368): undefined reference to `SystemCoreClock'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal_rcc.c.obj): in function `HAL_RCC_GetHCLKFreq':
[build] stm32h7xx_hal_rcc.c:(.text.HAL_RCC_GetHCLKFreq+0x54): undefined reference to `D1CorePrescTable'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: stm32h7xx_hal_rcc.c:(.text.HAL_RCC_GetHCLKFreq+0x58): undefined reference to `SystemD2Clock'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: stm32h7xx_hal_rcc.c:(.text.HAL_RCC_GetHCLKFreq+0x5c): undefined reference to `SystemCoreClock'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal_rcc.c.obj): in function `HAL_RCC_GetPCLK1Freq':
[build] stm32h7xx_hal_rcc.c:(.text.HAL_RCC_GetPCLK1Freq+0x28): undefined reference to `D1CorePrescTable'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal_rcc.c.obj): in function `HAL_RCC_GetPCLK2Freq':
[build] stm32h7xx_hal_rcc.c:(.text.HAL_RCC_GetPCLK2Freq+0x28): undefined reference to `D1CorePrescTable'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal_rcc_ex.c.obj): in function `HAL_RCCEx_GetD3PCLK1Freq':
[build] stm32h7xx_hal_rcc_ex.c:(.text.HAL_RCCEx_GetD3PCLK1Freq+0x28): undefined reference to `D1CorePrescTable'
[build] /Library/DaisyToolchain/0.2.0/arm/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: _deps/daisy-build/Drivers/libSTM32H7XX_HAL_DRIVER.a(stm32h7xx_hal_sai.c.obj): in function `SAI_Disable':
[build] stm32h7xx_hal_sai.c:(.text.SAI_Disable+0x68): undefined reference to `SystemCoreClock'
``` 

All other targets seem to be fine with GCC and all targets including Debug are fine with LLVM. Looks like GCC optimizes out some references from static lib because they are dependent on 

```
#ifndef BOOT_APP
SystemInit();
#endif
```

in `startup_stm32h750xx.c` or something like that 🙂

stellar-aria:
@asavartsov thank you so much for testing this all out! It looks like because the Makefile was building everything as one large static blob it didn't care so much about the references, but system_stm32h7xx.c actually should be compiled as part of the CMSIS Device library.

I've moved it there and changed the library type to static, as well as adding the compile definitions if the `DAISY_STORAGE` variable is set to anything other than `flash`.

This should fix the issues, please let me know if you have any more problems!

cmake/DaisyProject.cmake
```diff
@@ -0,0 +1,95 @@
+set(DAISY_STORAGE flash CACHE STRING "Select the storage destination of the executable.")
+
+if(DEFINED CUSTOM_LINKER_SCRIPT)
+    set(LINKER_SCRIPT ${CUSTOM_LINKER_SCRIPT})
+else()
+    cmake_path(SET LINKER_SCRIPT NORMALIZE ${CMAKE_CURRENT_LIST_DIR}/../core/STM32H750IB_${DAISY_STORAGE}.lds)
+
+    if(NOT DAISY_STORAGE STREQUAL "flash")
+        target_compile_features(daisy PRIVATE BOOT_APP)
```
asavartsov, line 9:
```suggestion
        target_compile_definitions(daisy PRIVATE BOOT_APP)
```

```
CMake Error at build/_deps/daisy-src/cmake/DaisyProject.cmake:9 (target_compile_features):
  target_compile_features specified unknown feature "BOOT_APP" for target
  "daisy".
Call Stack (most recent call first):
  CMakeLists.txt:39 (include)
```

asavartsov:
@stellar-aria I am really sorry if I annoy you too much! I did a little bit more testing and I have some more comments:

1. I tested build with all targets and different optimization options. `FIRMWARE_*_OPT_LEVEL` does not seem to affect libDaisy itself. Also if I build my project in Debug or Release, libDaisy sources compiled without any optimization flags at all (also no ggdb3 for Debug and no -DNDEBUG for Release).
Produced ninja FLAGS for daisy files:

Release: 
```
-std=gnu++14 [skip] -Wall
```

Debug: 
```
-std=gnu++14 [skip] -Wall
```

MinSizeRel:
```
-Os -DNDEBUG -std=gnu++14 [skip] -Wall
```

RelWithDebInfo:
```
-O2 -g -DNDEBUG -std=gnu++14 [skip] -Wall
```

I does not look like a desired behaviour... 

2. `link_libraries(semihost)` may be a little bit too bold [here](https://github.com/electro-smith/libDaisy/pull/615/files#diff-7bc51b1833030391040411da5b0b2de37e42b0f9ce569e1d963c6123975bee4eR28), it would be very hard (if even possible) to override in the parent project. For GNU toolchain you have `set(CMAKE_EXE_LINKER_FLAGS "--specs=nano.specs --specs=nosys.specs" CACHE INTERNAL "Linker options")` which would allow overriding, may be it is worth to have something similar for LLVM to be on par? By the way, I can confirm that semihosting works like a charm with LLVM.

3. I tested ArmGNUToolchain 13.2.Rel1 and LLVMEmbeddedToolchainForArm 18.0.0. 

GNU seems to be working fine except I could not make semihosting to work.

LLVM not great news. I started to test different peripherals and I stumbled into hw.StartAudio causing hard fault with `BFARVALID` after the code enters `SaiHandle::Impl::StartDmaTransfer` if libDaisy is an optimized build. I can confirm at least `-Og`, `-O1`, `-O3` and `-Os` triggers that rendering audio useless if compiled with LLVM toolchain. With `-O0` audio starts fine and callback works. I am not sure if compiler omits something or it is CPU runtime optimization error or both. I am actively debugging it though I am not a real embedded dev 😭 but I hope I will figure it out and fix it separately. I know that it is not related to CMake support at all but more like an info for those who will try the same after this will be merged.

stellar-aria:
@asavartsov no worries! I'd rather get this all worked out in the PR than have to deal with it in mainline.

1) yeah the current setup will not propagate the build flags. ~~I'll need to think on this one. I know what's causing it but it might be tricky to work around.~~ Figured it out. I've added a new patch.

2) `semihost` shouldn't be there, that's a whoops from my experiments

3) I have _no_ idea what's going on there tbh. This does sound like a bug in the LLVM optimization passes which is something I have 0 experience debugging. If you're using Cortex-Debug or something like Segger's  Ozone, you can step-debug the assembly output of the problem function/object file. Doing that and manually checking your registers might be the key here. Also an A/B comparison of the output object files and/or going through and manually enabling the -Og optimizations one by one to find which one is the problem.

stellar-aria:
Yeah considering how heavily it's used in certain places of libDaisy, and even shown in the examples, I think that's a reasonable solution.

Ideally we could convert them to a static `hang()` function or something that's just `while(1) { asm("nop"); }` since that doesn't get utterly destroyed by the optimization passes, but that's a separate issue.

stellar-aria:
@beserge @stephenhensley any ideas when/how this might get merged? Are there any edits that need to happen for it to be done?

beserge:
Checked out and built, everything appears as it should be. I'll defer to @stephenhensley on the merge timing, but looks good to go on my end! One thing to note, I had to update my CMake install, my out of date version was giving me some trouble, but once that was done it was smooth sailing.

stellar-aria:
Had to do manual merge resolution, since usbh_midi.c was already listed in the new refactored CMakeLists.txt. However, there also seems to be some issues with building on LLVM (on the CI side) now, so I'm going to poke about at that to see what's wrong.

stellar-aria:
Alright, everything's working as it should now.

Had to remove support for older versions of the LLVM Embedded Toolchain for ARM due to major breaking changes where the ARMv7E-M and ARMv7-M runtime libraries were merged, but considering we're only merging this now and 19.1.1 is the most current as of this comment, I think that's an acceptable solution.

stephenhensley:
@stellar-aria Sounds good! I've been doing a little review/testing on this this week; I'll re-sync everything locally, and continue to test a bit more.

So far, so good. Since it's getting a bit late today, and I'd like to test converting a few projects with the latest version, I'll probably end up merging this on Monday unless any issues arise.

stephenhensley:
So I am running into one issue that I didn't realize initially. 
I've verified that I can get this to happen when building on Mac OS and Windows.

On Mac OS, I've built with ARM GCC 10.3-2021.10 (10.3.1 20210824 release)
On Windows I've built with ARM GCC 10-2020-q4-major (10.2.1 20201103 (release))

In both cases, programs built seem to run properly after being programmed to the Daisy.
After a reset, or a full power cycle, it looks like the program faults immediately from the Reset_Handler. 

I've used the `DaisyBlinkProject` repo using the `FetchContent` feature, as well as modified the GPIO_Input example within the repo, and had the same experience.

CMakeLists.txt:

```CMake
cmake_minimum_required(VERSION 3.26)
cmake_policy(SET CMP0048 NEW)

include(FetchContent)
FetchContent_Declare(daisy
    GIT_REPOSITORY https://github.com/stellar-aria/libDaisy
    GIT_TAG 5086cc0
)
FetchContent_MakeAvailable(daisy)

project(daisyblink VERSION 0.0.1)
set(FIRMWARE_NAME daisyblink)
set(FIRMWARE_SOURCES
    ${CMAKE_CURRENT_LIST_DIR}/Blink/main.cpp
)
set(DAISY_GENERATE_BIN ON)

include(DaisyProject)

target_include_directories(${FIRMWARE_NAME} PRIVATE include)
```

main.cpp

```c++
#include "daisy_seed.h"

using namespace daisy;

DaisySeed hardware;

int main()
{
    hardware.Init();

    while (true)
    {
        hardware.SetLed(true);
        hardware.DelayMs(500);
        hardware.SetLed(false);
        hardware.DelayMs(500);
    }
}
```

---

Built with the following:

```
cmake -B build -G Ninja -D CMAKE_BUILD_TYPE=Release

cmake --build build --target all
```

---

I vaguely recall running into this a long time ago (possibly an issue with LTO, or something), but it was several years ago.

I'll retry with different optimization, etc. and see if I can narrow down the issue, but it's also possible I'm just overlooking something obvious.

stellar-aria:
@stephenhensley I've used your test case and it compiles and works (uploaded via the Daisy web programmer) with GCC 13.2.1 (see attached files) even after a cold-boot and power cycle, so next I'm going to try with 10-2020-q4.
[daisyblink.zip](https://github.com/user-attachments/files/17369122/daisyblink.zip)

stellar-aria:
I can confirm the behavior on 10-2020-q4, I'm going to check versions upwards to see where it resolves

stellar-aria:
Issue is present in 11.3.1 and resolved in 12.2.1

stephenhensley:
@stellar-aria thanks for stepping through that, and finding where it improved! -- I'll continue testing with a newer compiler.

Aside from this issue everything seems great so far. Last things I want to check are bootloader/linker script related, but I don't anticipate any issues.

stellar-aria:
yeah I'm not sure where this is going wrong. It seems like there might be a problem with linker stuff due to it just failing in the reset handler. For fun, here's a comparison between 11.3 and 12.2's first few bytes of the bin file. Unfortunately, I've never been good at decoding hex by hand.
image

asavartsov:
TLDR: Remove `-fno-builtin` from `stm32h750xx.cmake`.

ARM `GCC 10-2020-q4-major` with example above in Release configuration (`-O3`) generates floating point instructions (`vldr`, `vstmia`)  for the memory copy loop (😮) in the `Reset_Hanlder()` that is running before FPU enabled at `SystemInit()`. It causes a NOCP hard fault. 

image

GCC will usually optimize this loop to builtin memset but CMake build has `-fno-builtin` flag set (which usage is [strongly discouraged](https://github.com/ARM-software/CMSIS-DSP/blob/3a04f817a4380c00ff51f9bcdc6c4e4bb7a18b80/README.md#options-to-avoid) anyway). This flag is not set in any of Makefiles, this is why it always worked before.

stellar-aria:
Thank you so much for figuring that out @asavartsov. It looks like that was added in one of the first few refactor commits I was doing, probably pulled from somewhere else originally, and it's been so long I have no idea what prompted it being there or what I was thinking 😅 

It looks like without it, there's now some size changes in some of the binaries for clang, which is preventing them from fully building, so some of the examples might have to be prodded to work properly or removed from the test build

stellar-aria:
The culprit seems to be that without `-fno-builtin`, clang is placing some structures in the `.data` section that were previously in `.bss`, specifically the global variables in the main file (hw, midi, event_log).
image

stephenhensley:
> The culprit seems to be that without -fno-builtin, clang is placing some structures in the .data section that were previously in .bss, specifically the global variables in the main file (hw, midi, event_log).

Very strange indeed. Good catch.

@asavartsov thanks for digging in, and clarifying what was going on a bit more!

---

I think for the time being, we can except any of the currently included examples that are too large in order to get an initial version of this merged/passing CI again.

Another, slightly more medium-term, solution would be to build those examples with the bootloader (minding that at least the SDMMC example would require a custom linker so the FIL objects, etc. can be in AXI SRAM). However, I think coverage for all examples w/ CMake can be a goal we work toward rather than requiring it as a part of this PR.

I haven't personally run into any further issues yet, but I haven't had a ton more time for further testing. This also doesn't really have much effect on the (current) primary Make-based build system that seems unaffected by any of the llvm/cmake build issues we've discussed in the last few days.

Let me know what you think, and we can go from there.

stellar-aria:
@stephenhensley Yeah, I'm in agreement. Investigation on my end has indicated that this behavior is actually Clang/LLVM _more strictly_ conforming to the C++ standard than GCC does, which moves this from a codebase error to something more complicated. I might poke some LLVM devs and ask if they have any suggestions for how to go about achieving the behavior we want (large array-like globals in .bss).

In the meantime, I've reduced the Clang CI's build tasks from all of the examples down to just BootloaderBlink, which should at least verify that the CI _works_, even if not for all of our general GCC test cases. Meanwhile the GCC stuff is all up to date with the current collection of examples.

I'd say this is probably good to merge at this point, as long as you find no other issues. Bringing up conformance between the two compilers can definitely be another PR, if at all.

asavartsov:
In this particular case one could (and probably should) put MIDII/FIFO objects (which will take buffers with them) in any other section of their choice outside of FLASH region manually. Or avoid globals and trust compiler but it would be a bit trickier if you need to interact with audio loop since you can’t pass anything (I would imagine at least some opaque pointer) to it. So means to override this more or less exist for end user.

stephenhensley:
Alright! I'm signing off. 
Agreed that the CI validation is a good initial step, and full build coverage can be worked on over time.

Thanks again, @stellar-aria for the tremendous PR! And sorry for how long it took to get it merged.
I'm trying to spend a little bit of time on the library each week now. So things should move a little more consistently.
This was one of the last things to get in (aside from some QSPI/bootloader stuff I'll squeeze in next week) before the we do a v8 release with all of the recent merges.

PR #619, opened 2024-03-02
4 new fonts for OLED display added

atoav:
Adds 4 new fonts to oled_fonts:

- 4x6
- 4x8
- 5x8
- 6x7

These have been generated with the tool described in [this thread](https://forum.electro-smith.com/t/wrote-a-cli-tool-to-help-create-fonts-for-daisy/4990)

Ah, and the designs are by me (but lets face it: there are only so many ways to design fonts at such small sizes), I didn't include the "source" pxf files as this seemed a little bit excessive and I did not know where to put them.


beserge:
Works great! Tested using https://github.com/electro-smith/DaisyExamples/pull/300

PR #621, opened 2024-03-11
Move midi message text mapper into MidiEvent

GregBurns:
There was a comment in the UART and USB examples about moving the midi message type string mapping. I needed this function when debugging my application so decided now was a good time to move it to a shared location.


beserge:
Much cleaner, thanks for the contribution!

PR #623, opened 2024-04-04
Remove WM8731 TODO comment

jacobvosmaer:
I was reading the WM8731 audio codec setup code and I got confused by a comment which suggested Daisy always uses the codec with a 48kHz sample rate. After several hours of reading data sheets I figured out what is going on here: the WM8731 has built-in sample rate conversion and the nominal "48kHz" setting disables the built-in conversion.

This commit updates the comment to explain what we are actually doing here.


beserge:
Looks good, nice catch!

PR #624, opened 2024-04-28
Add SSD1307Driver

AndrewCapon:
Added code for SSD1307 displays.

As pimpl is being used I wasn't sure if you wanted any kind of inheritance so SSD1307Driver is a full implementation. 

Really only the Init() is any different to the SSD130x


stephenhensley:
Having a full implementation is fine for now.
There are a few other new display drivers being added as well. So we may be able to look at the full set some time down the road and see if there's a decent way to reuse any commonly shared code between them whether that's through some composition or inheritance.

Thanks for the contribution!

PR #625, opened 2024-04-29
Add drivers for SSD1327 and the SSD1351.

a7v7:
Add color display driver and OLED color display driver / template. These are closely related to the display and OLED display driver. Updated daisy.h to include the color driver and the OLED color driver. The hardware interface is identical to the black / white driver.

The SSD1327 driver is configured for 128x128. It supports 16 gray scale level. Each pixel is 4-bits.
The SSD1351 driver is configured for 128x128. It supports 64k color. Each pixel is 16-bits (565 color).

The main addition to the black / white driver is the concept of a foreground / background color. All other calls remain the same.
These device drivers are a v1 status. I want to add image processing and change the SPI transaction to DMA.

These drivers were tested with these two OLED displays:
SSD1327: [WaveShare OLED](https://www.waveshare.com/product/displays/oled/oled-3/1.5inch-oled-module.htm)
SSD1351: [WaveShare OLED](https://www.waveshare.com/product/displays/oled/oled-3/1.5inch-rgb-oled-module.htm)

### DIFF
--- a/src/daisy.h
+++ b/src/daisy.h
@@ -37,6 +37,8 @@
 #include "per/rng.h"
 #include "hid/disp/display.h"
 #include "hid/disp/oled_display.h"
+#include "hid/disp/color_display.h"
+#include "hid/disp/oled_color_display.h"
 #include "hid/disp/graphics_common.h"
 #include "hid/wavplayer.h"
 #include "hid/led.h"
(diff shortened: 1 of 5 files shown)


stephenhensley:
Awesome! Thanks for the contribution!

PR #627, opened 2024-05-03
Add usbh_midi.c to CMakeLists.txt

asavartsov:
#611 added `Middlewares/ST/STM32_USB_Host_Library/Class/MIDI/Src/usbh_midi.c`  to `Makefile` but not to `CMakeLists.txt`.

Let's maybe have it because without it you can't really have USB MIDI working (even if you don't use host option) using provided `CMakeLists.txt` because of missing symbols during link time.


stephenhensley:
thanks for the contribution!

PR #628, opened 2024-05-04
Fix I2C4 not working

AndrewCapon:
Use GPIO_AF6_I2C4 for alternate function


beserge:
Unable to confirm this works. Tested on the daisy field, replaced I2C1 with I2C4, which should work since they share pins, but it stops working. More digging may be needed.

beserge:
Sorry for the back and forth, I was able to retest and confirm this is working. Thanks for your contribution!

PR #630, opened 2024-05-17
Update daisy_patch_sm.cpp  DacHandle::BufferState::DISABLED;

IC-Alchemy:
Specific to the  Patch_SM, only changed one line of code.

` DacHandle::BufferState::DISABLED;`

Increases DAC output range to 0.0v - VREF
Quiet a few people believe that this cannot be done it seems.

image
```

Page 163
https://www.st.com/resource/en/datasheet/stm32h750ib.pdf

### DIFF
--- a/src/daisy_patch_sm.cpp
+++ b/src/daisy_patch_sm.cpp
@@ -181,7 +181,7 @@ namespace patch_sm
         dac_config.bitdepth = DacHandle::BitDepth::
             BITS_12; /**< Sets the output value to 0-4095 */
         dac_config.chn               = DacHandle::Channel::BOTH;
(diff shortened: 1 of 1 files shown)


stephenhensley:
Sorry for the near-two-year delay on getting to this.
This definitely results in something more similar to what a user would expect, and given the PatchSM has an output gain stage for the DAC outs, this makes total sense.

I took a few measurements for posterity:

| DAC Output | buffer ENABLED | buffer DISABLED |
| --- | --- | --- | 
| min output `0x000` | 0.045V | 0.009V |
| max output `0xfff` | 4.943V | 4.980V |

Thanks for the contribution!

PR #643, opened 2024-09-23
Feature: Midi - make it possible to handle 'SystemRealTime' events synchronously

HeckHeckHeckHeck:
Sync more precisely to a Midi clock signal.

### DIFF
--- a/src/hid/midi.h
+++ b/src/hid/midi.h
@@ -5,6 +5,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <algorithm>
+#include <functional>
 #include "per/uart.h"
 #include "util/ringbuffer.h"
 #include "util/FIFO.h"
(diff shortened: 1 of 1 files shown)


stephenhensley:
I merged the more passive realtime-related PR first so that I could test this on top of it, and make sure everything seemed to be consistent (mostly since this is opt-in, and that one may resolve some issues people were running into).

Just finished testing this out a bit and comparing between the two modes, and everything seems to work great in the few cases I tested it.

I didn't make any practical programs that would take advantage of the improvement yet (though I do have a few projects that I think could benefit from this update). 
However, I have some changes to the MIDI_UART_Input example that allow for selecting between the realtime callback or not, and log timestamps of when the message was handled by the application, and the source (queue, or real time callback). I used this modified example to do some testing on the hardware and validate the changes.
I'll add that once this is merged so it can act as a sort-of demo to using the new feature.

Thanks again for the contribution, @HeckHeckHeckHeck!

@TomFrost It looks like the changes in #643 implement the fix that you had staged. So everything should be pretty solid in the System Realtime now.

PR #645, opened 2024-09-23
Build: Core/Makefile - support c++ sources with suffix '.cc' or let t…

HeckHeckHeckHeck:
…he user choose by defining USER_SUFFIX_CXX

My implementation is not pretty, but it works (for me at least)
Maybe a makefile ninja can improve it

The intent and motivation should be clear. 
Currently users are forced to use the .cpp suffix. (which is very uncommon in linux/unix land at least)

As the TODO comment states, a nice build system config interface description in the form of a Makefile.conf or so would be a sweet next step.


HeckHeckHeckHeck:
Replaced by/Contained in: https://github.com/electro-smith/libDaisy/pull/647

PR #650, opened 2024-10-08
Fix: New OLEDs (SSD1327 & SSD1351) were still using old GPIO

### LINKED ISSUE #649: SSD1351 code cannot be compiled
The oled_ssd1351.h/cpp still uses the old (deprecated) gpio calls and definitions and no longer can be compiled after the latest libDaisy release.

stephenhensley:
updated oleds ssd1327 and ssd1351 to modern Pin/GPIO API

Out of the scope of this quick-fix, but it looks like the Transport class for all three of the SSD13xx drivers in the library are all identical. So a minor refactor would help reduce the opportunity for similar regressions in the future (as would examples for each driver).

### DIFF
--- a/src/dev/oled_ssd1327.h
+++ b/src/dev/oled_ssd1327.h
@@ -22,8 +22,8 @@ class SSD13274WireSpiTransport
         SpiHandle::Config spi_config;
         struct
         {
-            dsy_gpio_pin dc;    /**< & */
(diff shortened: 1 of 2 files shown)


stephenhensley:
I don't have either of these devices on hand, but I did modify the existing SSD130x example to use each display instead, and was able to compile without errors.

PR #657, opened 2024-11-23
Support for Trill capacitive sensors

### LINKED ISSUE #602: Feature: Trill sensor support
I was mentioned on the forum some ~2 years ago, but here an official ticket to request and track this.

We can already use Trill sensors using the Arduino library, but it would be great to also have them in libDaisy as a lot of projects use this rather than Arduino.

giuliomoro:
This is the "Linux" library (available [here](https://github.com/BelaPlatform/Trill-Linux)) wrapped into libDaisy. The low-level I2c I/O stuff is done here:  https://github.com/giuliomoro/libDaisy/blob/64c08470c079693cd9e2ab59cf2b85a5cff4727c/src/dev/trill/I2c.h

Contributions by me and @virvel .
Tested by @dromer and @virvel when it was based on 17ee39bac370b0948949be94fa7d0fcbe890ffd7.
Previous discussion https://github.com/giuliomoro/libDaisy/pull/1

Closes https://github.com/electro-smith/libDaisy/issues/602


virvel:
> OK it's now brought up to date with the latest upstream and I even figured out how to do the formatting. @virvel @dromer I'd appreaciate if you have a chance to test this again.

I can confirm that this works!

dromer:
@giuliomoro can you have a look at the comments? would be nice to move this forward!

giuliomoro:
@dromer I'd love to, but it's going to be a month till we ship the thing I am currently working on, before I can get back to this. 

I'd really want to get this merged swiftly and not add unnecessary burdens, but I want to have a look ahead for maintainability purposes.

In general I'd really wish I could just drop the upstream files in here, with minimal wrapping modifications, rather than having to rename methods and variables to conform to a style. I appreciate that daisy strives to present a uniform-ish interface, but if that means that all changes for external libraries (such as this one, not sure if you have others or plan to) will require more effort to be backported, it may be an obstacle to keeping the library up to date.

So, wrt https://github.com/electro-smith/libDaisy/pull/657#issuecomment-3946610486 @stephenhensley , would it be OK to only do the numbered list items and not the bullet points?

An alternative is to provide a small wrapping class (e.g.: `DaisyTrill`?) that complies with everything you say. It would privately include Trill.h and hold a pointer to a Trill object and expose a minimal set of methods and a `Trill* getTrill()` method for those interested in including `Trill.h` directly and accessing its more advanced functions. With such a wrapper in place, the library update process would be:
- get upstream Trill library
- pass it through clang-format
- commit

Or, even better, the most maintainable solution:
- include the upstream Trill repo as a submodule
- to update, simply update the submodule hash and commit

dromer:
The wrapping class actually sounds like a pretty good idea as it would give the most consistent user-facing API while keeping the Trill library separate and easy to update (a submodule sounds good!)

Of course lets see what @stephenhensley has to say ;)

stephenhensley:
@giuliomoro thanks for your reply!

I'll have to think a bit on the overall plan, but my initial thoughts are that the minimum criteria from my first comment:

> 1. The Trill object should be added to the daisy namespace instead of being global
> 2. I'd suggest that we also wrap most/all of the internals that have been added to support the Trill object in a daisy::trill namespace (this is to help avoid things like trill's I2c from being included in the global or daisy namespace and causing any confusion.
> 3. the /src/dev/trill/Trill.h file should be included in daisy.h so user's don't need to manually include the path (the examples could then be updated as well)

would still be things I'd consider requirements for this to be merged, with all other style-conformance/uniformity being optional.
The reason for these is primarily to prevent overlap with users' app code, and collisions within the rest of libDaisy with generic naming of things like `I2c`, etc.

My initial impression of this contribution was to provide (with the linux driver as a starting point) an API for accessing trill sensors that would eventually feel consistent with the rest of libDaisy, and not just a bundling the existing trill-linux sources into libDaisy (similar to how CMSIS DSP is currently included in the library).

With that said, I am open to the latter approach, whether that includes switching the directly copied files to be located in a submodule, or leaving modified copies within libDaisy (maybe more clearly in a separate third-party source directory). 
However, the Bela-trill repo is not very active, and has not been updated in the last few years. So I personally don't think there is a ton of value in making upstream-syncing a priority.

PR #663, opened 2025-04-02
Fixes wav files written with WavWriter.h to be complete

willemOH:
Before the transfer buffer was not flushed so wav header files would say the file was a certain size but actually it would be smaller because the full data was not written to the wav file. This would result in perfect loops being recalled imperfectly. See https://github.com/willemOH/daisy_looper/tree/main commits for debugging which proves this behavior.

### DIFF
--- a/src/util/WavWriter.h
+++ b/src/util/WavWriter.h
@@ -143,8 +143,14 @@ class WavWriter
     {
         unsigned int bw = 0;
         recording_      = false;
-        // We _should_ flush whatever's left in the transfer buff
-        // TODO: that.
+
+        // Flush remaining data in the transfer buffer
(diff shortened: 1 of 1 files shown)


nwesem:
hi @willemOH i found this PR yesterday evening and it seems that it solves the issue i'm facing. thanks a lot for this. do you however have a running code snippet to write a collected audio buffer to sd card? I already tested it using my implementation and the results look better but I somehow get distorted audio and can't reason why

stephenhensley:
Thanks for the much needed contribution! 😄 

This looks good to me.
I'll do some testing, and get this merged some time next week!

stephenhensley:
Wrote a quick example to test this (offline rendering a 1s audio file), and can confirm the full file now gets written.

I'll add the example in a separate PR.

Issue #622, opened 2024-03-29
I2C address representations are inconsistent


simon-zumbrunnen:
I used TransmitBlocking and everything worked as expected. When I tried to use WriteDataAtAddress, because it fit my use case better, nothing worked anymore. After I checked the signals with a scope I found out that TransmitBlocking shifts the address to the left while WriteDataAtAddress does not. This is inconsistent. I think shifting is the more intuitive approach.

PR #668, opened 2025-05-14
Left-shift addresses in Read/WriteDataAtAddress

cvpines:
Left-shift the I2C device addresses passed into Read/WriteDataAtAddress, for consistency with the other I2C functions.

See #622 for details.

Note that this change is not backwards-compatible!


cvpines:
Will update existing uses of this library then un-draft

cvpines:
Updated all existing uses of affected methods (PCM3060, MCP23x17 drivers)

stephenhensley:
Confirming that I checked the PCM3060 via the Patch SM Board.
I am unable to check the MAX device that was affected, but the change looks good.

PR #669, opened 2025-05-15
added functions for reading mpr121 capacitive sensor

### LINKED ISSUE #651: MPR121 class FilteredData() returns wrong data
see thread here: 
[https://forum.electro-smith.com/t/mpr121-filtered-data-reading-wrong/5921](https://forum.electro-smith.com/t/mpr121-filtered-data-reading-wrong/5921 )

Matteocaroleo:
Hi, we at Politek (the audio engineering team at Politecnico di Torino) have found the problem in the libDaisy library and propose a solution.

You can check us out at https://politek.polito.it/
### Problem
In the files libDaisy/src/per/i2c.h and libDaisy/src/per/i2c.cpp to read from an I2C device the function ReceiveBlocking(...) is used.
Looking at the i2c.cpp file, at line 346, in its function definition the hal function HAL_I2C_Master_Receive (...) is used. The problem is that this hal function doesn't comply with the I2C reading protocol defined in the MPR121 datasheet, resulting in a wrong read.

### Solution
We found that the hal function HAL_I2C_Mem_Read (...) implements the correct procedure to read from the MPR121, so we added a new function inside the i2c.cpp and inside the src/dev/mpr121.h files to correctly use it.
Using the function FilteredData() now returns the correct values of capacitance reading from the mpr121 device.

Closes #651

### DIFF
--- a/src/dev/mpr121.h
+++ b/src/dev/mpr121.h
@@ -72,7 +72,17 @@ class Mpr121I2CTransport
         return I2CHandle::Result::OK
                != i2c_.ReceiveBlocking(config_.dev_addr, data, size, 10);
     }
-
+    bool ReadMem(uint8_t *data, uint8_t memAddress, uint16_t memSize)
+    {
+       return I2CHandle::Result::OK
+              != i2c_.Mem_read(
+                   config_.dev_addr, 
+                   memAddress,
+                   1,
+                   data,
+                   memSize,
+                   10);
+   }
   private:
     I2CHandle i2c_;
     Config    config_;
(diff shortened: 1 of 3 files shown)


cvpines:
Hi @Matteocaroleo, `HAL_I2C_Mem_Read` is already implemented via `I2CHandle::ReadDataAtAddress`. If you look at the source for that method, you will see that it is implemented the same way as method this PR adds.

The only difference is that `I2CHandle::ReadDataAtAddress` does not currently shift the address left before passing it, but #668 adds that change.

Matteocaroleo:
You're right, so the `ReadDataAtAddress()` function can be used inside the mpr121 class instead of the one made to make it work, since it's the same anyway. I will do that and make another PR

PR #672, opened 2025-06-16
Wav Writer Example

stephenhensley:
Wrote this brief example to test #663

It offline renders a 220Hz sine wave to a 16-bit 48kHz audio file.


stephenhensley:
I could have sworn the style checks passed on the related PR, but I guess those fixes are getting applied here 🤷

PR #678, opened 2025-10-08
Add missing volatile to status flags to prevent "infinite" loop after optimization

TheSlowGrowth:
To prevent the while loops from being optimized into very long loops.

Without this change, loops such as this one in `SD_read()` (`sd_diskio.c:154`) get "optimized" into 30s loops, when the `(ReadStatus == 0)` check gets optimized out:
```C++
        while((ReadStatus == 0) && ((HAL_GetTick() - timeout) < SD_TIMEOUT)) {}
```
This can be observed with arm-gcc-none-eabi 10 2021-q4-major using `DaisyProject.cmake`
This makes the SD unusable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+- SDMMC: fix an issue where status flags were optimized out, resulting in very long while loops
+
 ## v8.0.0
 
 ### Features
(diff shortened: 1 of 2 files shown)


stephenhensley:
Woah, fascinating.

I've hit this loop a few times before, but never reliably.
I have almost entirely used the Make builds, so maybe some flag set there causes this to be more intermittent than with the CMake build.

In either case, nice catch. Thanks for the update!
I'll give this a quick test before the end of the week, and get it merged!

PR #693, opened 2026-02-24
Fix: MIDI parser — System Real-Time no longer corrupts running status

### LINKED ISSUE #422: MIDI - Clock messages may interfere with running status implementation
makingsoundmachines on Slack reported that when using the TR-8,  note messages are getting lost, between clock messages.

> . . .the Roland TR8 sends F8 (so clock tick) messages in the middle of these; e.g. 90 24 FF .. then 24 00 F8 .. so the status hasn't changed but there are clock ticks somewhere in between.

I have a feeling our implementation interprets clock messages as status bytes, and are potentially clearing the running status byte when it should persist.

Worth looking into.

### LINKED ISSUE #665: Midi NoteOn Velocity 0 wrong channel
I’ve got midi working fine with devices that use NoteOn and NoteOff message types. However, when I use a device that sends NoteOn Velocity 0 instead of NoteOff, I have problems. I see there is a if statement in midi_parser.cpp on line 143 that appears to handle this case:
```

//velocity 0 NoteOns are NoteOffs
                if(running_status_ == NoteOn && incoming_message_.data[1] == 0)
                {
                    incoming_message_.type = NoteOff;
                }

```
However, when I send midi data that uses NoteOn Velocity 0 for note offs, I find that, while they do get parsed as NoteOffs, I get the wrong channel on NoteOff messages from my program:
```
// Listen to MIDI for new changes
         midi.Listen();

         // When there are messages waiting in the queue...
         while (midi.HasEvents()) {
             // Pull the oldest one from the list...
             auto msg = midi.PopEvent();
             /* if(msg.channel != midiChannel){
                break;
             } */
             switch (msg.type) {
                 case NoteOn: {
                    auto note_msg = msg.AsNoteOn();
                    voiceManager.OnNoteOn(note_msg.note, note_msg.velocity);  
                    mLog(MIDI, "on channel: %d", note_msg.channel);
                    break;
                 }
                 case NoteOff:{
                    auto note_msg = msg.AsNoteOff();
                    voiceManager.OnNoteOff(note_msg.note, note_msg.velocity);
                    mLog(MIDI, "off channel: %d", note_msg.channel);
                    break;
                 }
                 
                 default:
                     break;
             }
         }
```
You can see I ran into this when I tried to add a midi channel filter (commented out). With the usual midi device (proper NoteOffs sent) I would get this output:
```
on channel: 3
off channel: 3
on channel: 3
off channel: 3
on channel: 3
off channel: 3
```
etc.

But with a velocity 0 noteons I get this:
```
on channel: 3
off channel: 8
on channel: 8
off channel: 8
on channel: 3
off channel: 8
on channel: 3
off channel: 8
on channel: 8
off channel: 8
on channel: 3
off channel: 8
on channel: 3
off channel: 8
on channel: 8
```
You’ll notice that not only do the off notes have the wrong channel, but something gets mixed up enough to occasionally assign the wrong channel to note ons too. This seems like a bug in libDaisy centering around `midi_parser.cpp`

### LINKED ISSUE #666: Midi In Clock not supported
Currently MidiEvent.h states that only 3-byte messages are handled properly at this time. Clock messages are 1-byte. 

I experienced how improperly they are handled when I tried to figure out for a day why my incoming midi notes channels were changing inexplicably. It was my device sending midi clock which corrupted the stream. 

Fix: support Clock In

sauloverissimo:
## Summary

System Real-Time bytes (0xF8–0xFF) can legally appear *anywhere* in a MIDI byte stream — even between data bytes of a multi-byte message — and must not affect the parser's running status. The current parser handles them inside the `ParserEmpty` switch case, which means they first trigger a state reset (`pstate_ = ParserEmpty`) and overwrite `incoming_message_` fields. This causes:

1. **Channel corruption** — a Timing Clock (0xF8) between NoteOn bytes shifts the channel to 8 (since `0xF8 & 0x0F = 8`)
2. **Lost messages** — the in-progress multi-byte message is destroyed
3. **SysEx corruption** — RT bytes inside SysEx are stored as data instead of being dispatched

This is the root cause of #666 and #665.

## The fix

Move System Real-Time handling to the **very top** of `Parse()`, before any state modification:

```cpp
if(byte >= 0xF8)
{
    if(event_out != nullptr)
    {
        MidiEvent rt_event;
        rt_event.type    = SystemRealTime;
        rt_event.channel = 0;
        rt_event.srt_type = static_cast<SystemRealTimeType>(byte & kSystemRealTimeMask);
        *event_out = rt_event;
    }
    return true;
}
```

Uses a **stack-local** `MidiEvent` so `pstate_`, `incoming_message_`, and `running_status_` are completely untouched. The early return means RT bytes are transparent to the rest of the parser — exactly as the MIDI 1.0 spec requires.

## Tests

Added 4 new unit tests covering the scenarios that the existing `runningStatSysRealtime` test does not:

| Test | What it covers |
|------|---------------|
| `sysRealtimeMidMessage` | Clock bytes interleaved *within* a NoteOn (between status/data0/data1) |
| `sysRealtimeDuringSysEx` | Clock byte inside a SysEx message — must not corrupt SysEx data |
| `allSysRealtimeTypes` | All 8 RT types (0xF8–0xFF) parse correctly with channel=0 |
| `runningStatusPreservedAfterRealtime` | Running status NoteOn + RT + channel verification across multiple notes |

**All 154 tests pass** (including the existing `runningStatSysRealtime` test).

## Related

- Closes #666
- Closes #665
- Re-resolves #422 (the mid-message case was not covered by the original fix)
- Complements PR #643 (synchronous RT callback) — that PR adds a callback feature at the `MidiHandler` level, but the underlying parser bug still exists without this fix

### DIFF
--- a/src/hid/midi_parser.cpp
+++ b/src/hid/midi_parser.cpp
@@ -4,6 +4,24 @@ using namespace daisy;
 
 bool MidiParser::Parse(uint8_t byte, MidiEvent* event_out)
 {
+    // Handle System Real-Time (0xF8-0xFF) immediately, regardless of
+    // parser state. Per the MIDI spec, these single-byte messages can
+    // appear anywhere in the stream — even between data bytes of another
+    // message — and must NOT affect running status or parser state.
+    if(byte >= 0xF8)
+    {
+        if(event_out != nullptr)
+        {
+            MidiEvent rt_event;
+            rt_event.type    = SystemRealTime;
+            rt_event.channel = 0;
+            rt_event.srt_type
+                = static_cast<SystemRealTimeType>(byte & kSystemRealTimeMask);
+            *event_out = rt_event;
+        }
+        return true;
+    }
+
     // reset parser when status byte is received
     bool did_parse = false;
 
@@ -21,8 +39,6 @@ bool MidiParser::Parse(uint8_t byte, MidiEvent* event_out)
                 incoming_message_.channel = byte & kChannelMask;
                 incoming_message_.type
                     = static_cast<MidiMessageType>((byte & kMessageMask) >> 4);
-                if((byte & 0xF8) == 0xF8)
-                    incoming_message_.type = SystemRealTime;
 
                 // Validate, and move on.
                 if(incoming_message_.type < MessageLast)
@@ -50,20 +66,6 @@ bool MidiParser::Parse(uint8_t byte, MidiEvent* event_out)
                             }
                             did_parse = true;
                         }
-                    }
-                    else if(incoming_message_.type == SystemRealTime)
-                    {
-                        incoming_message_.srt_type
-                            = static_cast<SystemRealTimeType>(
-                                byte & kSystemRealTimeMask);
-
-                        //short circuit to start
-                        pstate_ = ParserEmpty;
-                        if(event_out != nullptr)
-                        {
-                            *event_out = incoming_message_;
-                        }
-                        did_parse = true;
                     }
                     else // Channel Voice or Channel Mode
                     {
--- a/tests/Midi_gtest.cpp
+++ b/tests/Midi_gtest.cpp
@@ -654,6 +654,151 @@ TEST_F(MidiTest, runningStatSysRealtime)
     EXPECT_EQ(noteon2.type, NoteOff);
 }
 
+/** System Real-Time bytes arriving between data bytes of a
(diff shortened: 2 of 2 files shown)


sauloverissimo:
Thanks @stephenhensley! Let me know if you need anything or if I should make any adjustments. 🌟

PR #698, opened 2026-04-07
Fix USBH_MIDI_Transmit: use DMA-safe buffer and serialize transfers

jpanneton:
The caller's buffer was passed directly to USBH_BulkSendData. On the STM32H750, this buffer may reside in DTCM or cached AXI SRAM, which are memory regions the USB OTG peripheral cannot reliably access. This caused the device to ACK but silently discard the payload, leading to OUT transfers sending corrupted data.

Additionally, rapid back-to-back calls would re-submit on the same OUT pipe before the previous transfer completed, since USBH_BulkSendData sets the URB state to URB_IDLE on submission, making it indistinguishable from "no transfer pending"

Changes to USBH_MIDI_Transmit:
- Copy TX data into the handle's txBuffer before sending
- Track in-flight state with a txActive flag and wait for URB_DONE from the previous transfer before submitting the next


stephenhensley:
Nice! An initial glance looks good to me!
I'll review this in more depth, and merge or provide any feedback in the next few weeks!

Thanks for the contribution! :)

PR #705, opened 2026-06-26
SAI - Allow both blocks to be set to Sync::SLAVE and still remain synchronous

stephenhensley:
This adds handling for two SAI blocks with no master (e.g. daisy following another daisy, etc.).

The immediate use case for this is to be able to set up a second Daisy as a DSP for a primary Daisy, where the secondary Daisy would be configured with both channels set to `Sync::SLAVE`.

Now, the SAI peripheral driver will detect that, and still set the `SaiInitTypeDef::Synchro` for the B slot to `SAI_ASYNCHRONOUS` so that they can share a callback in the typical fashion.

This has been working fine for me during an hour or so of testing with One daisy generating audio, sending it to the second to echo back to the first, and output on its SAI1 outputs.
However, I have noticed a few times where it seems like one of the streams is getting a bit off, before correcting itself.
I have a feeling this may be an issue with the SAI1 and SAI2 synchronization, which has come up a few times in the past, but has never been consistently reproducible.

I'll spend a little time looking at whether I can recreate that behavior, and possibly improve cross-SAI sync at the same time, but that may be something we take care of later if it isn't consistent.

---

I'll add a pair of examples, or share a new repo with programs for each side before merging this so others can use that as a jumping off point for multi-daisy projects 😄

### DIFF
--- a/src/per/sai.cpp
+++ b/src/per/sai.cpp
@@ -115,6 +115,9 @@ SaiHandle::Result SaiHandle::Impl::Init(const SaiHandle::Config& config)
             break;
         default: break;
     }
+    const bool both_blocks_slave = config.a_sync == Config::Sync::SLAVE
+                                   && config.b_sync == Config::Sync::SLAVE;
+
     // Audio Mode A
     if(config.a_sync == Config::Sync::MASTER)
     {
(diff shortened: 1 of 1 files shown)


stephenhensley:
> I'll spend a little time looking at whether I can recreate that behavior, and possibly improve cross-SAI sync at the same time, but that may be something we take care of later if it isn't consistent.

I tested a bit more, and haven't seen this happen again yet.

However, I did look a little bit, and there are both ways at syncing the SAIs to one another, and the pin for "I2S_CKIN" is pin D3 on the seed. 
So reconfiguring the clock to use MCLK from another daisy is also possible, though would require some tweaking of how we deal with the clock tree (SAI1 and SAI2 can even have their own clock inputs if that's ever useful).

Anyway, for now I've shared a [pair of demo programs in a repo](https://github.com/stephenhensley/Dual-Daisy-DSP) for now with this PR checked out for testing.

PR #706, opened 2026-07-25
Add TCA9534 I2C GPIO expander driver and example.

fmulassano:
Header-only device driver following the Mcp23x17 pattern, with both owned-bus and shared-bus Init for multi-device I2C boards. Includes example and changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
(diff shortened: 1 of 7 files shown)


stephenhensley:
Looks good to me!
I don't have one of these devices handy. So I'm unable to actually run your example, but it looks fine, and is a nice clean example.

Thanks for the contribution!

PR #717, opened 2026-09-23
Seed3 DevKits

stephenhensley:
Adds support for the upcoming Seed3 DevKits, and updates default library build settings to use C++17 instead of C++14.

These are similar to the wrappers for the kickstarter products (pod, patch, petal, field) with a few notable differences in structure:

Each class is simply named, `Hardware` and scoped within a namespace for that devkit (e.g. `desktop_devkit::Hardware`)
This is to make the process of moving firmware from one platform to another require a few less edits, and match up with common patterns in custom HW projects.

The API for each devkit is much more minimal. The common methods (e.g. `Init()`, `UpdateAllControls()`) are still provided, but there is an expectation that members (e.g. pots, switches, etc.) will be accessed directly. This reduces the code footprint, and again, makes adapting programs from the devkits to custom hardware require fewer changes and/or boilerplate.

The library was updated to build with C++17 (we may bump it up to C++20 later, as well).
This has been something requested for a few language features.
In the case of the devkits, the only c++17 feature being used are for the inline constexpr values for the quantity of pots, etc. (e.g. `inline static constexpr size_t kNumPots = 8;`). 
This, in my opinion, is the most legible way to have these scoped within the class while still having the value obvious to users/readers.

---

Accompanying the new hardware support classes, are a few new examples:

- Desktop-DevKit-UI-Template: Uses the libDaisy UI framework, and external USB bootloader. Prints many inputs from the UiPage, and others (CVs, gates, etc.) read from the audio callback over USB serial.
- Eurorack-DevKit-Template: Prints all inputs over USB serial, blinks LEDs, writes waveforms to DAC (CV outputs), and passes MIDI Note messages from MIDI input to MIDI output.
- Pedal-DevKit-Template: Prints all inputs over USB serial, blinks LEDs, and passes MIDI Note messages from MIDI input to MIDI output.


stephenhensley:
I'll look into the (seemingly unrelated?) cmake-gcc build error, and see if it has anything to do with the change to C++17.

If it does, I'll separate that version bump from this PR, and we can revisit that later.

stephenhensley:
Okay -- something weird w/ the C++17 was seeming to cause a build error during the build-cmake-gcc build with 9-2020-q2.

Also, since the dev kits are all namespaced, it is a bit cleaner to use the consts as: `daisy::pedal_devkit::kNumPots` than previously, `daisy::pedal_devkit::Hardware::kNumPots`.

So I've moved those back to just being namespaced constexprs, and i'll look into doing a C++ standard update sometime later.

